PR #5, opened 2015-09-04
Replaced binary operators with intrinsics in AVX2 and SSSE3 codecs, proper check of CPU's AVX2 feature support

zenden2k:
I have created a new pull request to keep history clean.

### DIFF
--- a/lib/codec_choose.c
+++ b/lib/codec_choose.c
@@ -1,15 +1,48 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stddef.h>
+#include <stdint.h>
 
-#if __x86_64__ || __i386__
-#include <cpuid.h>
+#if __x86_64__ || __i386__ || _M_X86 || _M_X64
+#ifdef _MSC_VER
+	#include <intrin.h>
+	#define __cpuid_count(__level, __count, __eax, __ebx, __ecx, __edx) \
+	{\
+		int info[4];\
+		__cpuidex(info, __level, __count);\
+		__eax = info[0];\
+		__ebx = info[1];\
+		__ecx = info[2];\
+		__edx = info[3];\
+	}
+	#define __cpuid(__level, __eax, __ebx, __ecx, __edx) __cpuid_count(__level, 0, __eax, __ebx, __ecx, __edx)
+#else
+	#include <cpuid.h>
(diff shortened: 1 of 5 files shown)


aklomp:
It's actually possible in GitHub to update a pull request by force-pushing over your branch. The pull request will be updated automatically.

lib/codec_choose.c
```diff
@@ -90,18 +117,41 @@ codec_choose_arm (struct codec *codec)
...
 	if (max_level >= 7) {
 		__cpuid_count(7, 0, eax, ebx, ecx, edx);
-		if (ebx & bit_AVX2) {
-			codec->enc = base64_stream_encode_avx2;
-			codec->dec = base64_stream_decode_avx2;
+
+		if ((ebx & bit_AVX2) && (ecx & bit_XSAVE_XRSTORE)) {
+			uint64_t xcr_mask;
+			xcr_mask = _xgetbv_32(0);
+			if (xcrFeatureMask & 0x6) {
```
aklomp: This causes a compilation error: `'xcrFeatureMask' undeclared`

zenden2k: Fixed

zenden2k:
Fixed and force-pushed

zenden2k:
Travis CI can't check AVX2?

aklomp:
No, the Travis virtual machines don't have AVX2 support. It might be possible to test AVX2 support by running the test suite under Intel SDE (an instruction set emulator).

Your changeset is starting to look good, and I will merge it when the last issues are worked out. (I might make some small changes to the codeflow later, but I will leave your contribution intact.) However, because it requires C99, I need to make some commits to this project to switch to C99 first. Working on it...

aklomp:
OK, this library is now C99 :)  Now that we're C99, I changed the return type of the `codec_*()` functions to `bool`.

I also updated the Moby Dick tests to include the plain and encoded data as inline strings. So the funky compile-as-binary-object stuff is gone, and the test should work on Windows.

zenden2k:
>No, the Travis virtual machines don't have AVX2 support.
Maybe it could at least check that compilation succeeds with AVX2 flag turned on.

aklomp:
Merged, thanks!

PR #6, opened 2015-09-14
Benchmark: now compiles on Windows

zenden2k:
Tested on MSVC and Mingw

PR #7, opened 2016-01-16
Add CMake project

### LINKED ISSUE #70: Is shared library supported?
I want to build the library as a shared library(may be called libbase64.so),any suggestions?

BurningEnlightenment:
Hi,
I just made a CMake script for the project, because I want to include it within another CMake project. The CMake script includes an install target (mostly for windows) and supports SIMD on the x86 platform (supporting NEON shouldn't be hard, but I don't need it). However I had to change `#include "config.h"` in order to support out of source tree builds. I hope you are ok with my solution.
I successfully built the library on windows with mingw-w64 gcc 4.8.2 x86_64 and Visual Studio 2015 x86/x86_64/ARM.

My script is CC0 licensed and the `TargetArch.cmake` is distributed under a 2-clause BSD license (please note that no one distributes a cmake script in binary form, so the second clause concerning binary distribution is more or less meaningless).

Fixes #70 

Cheers,
Henrik


BurningEnlightenment:
Well, actually I tested the modified makefile using the windows cmd and msys2/zsh and mingw :confused:.

aklomp:
Hi! Thanks a lot for this contribution! As you can tell from [the other open pull request](https://github.com/aklomp/base64/pull/6), this project is definitely in need of a good cross-platform build system. I think that CMake is probably the best candidate, and that your pull request is a good start. I'd like to work with you to tailor it to this project.

I really appreciate your contribution back to this project, and I understand that I have little ground to critique work done by others in their own time. That said, I have some comments, which I will place in the code. It would be nice if you could have a look at them, but you are, of course, under no obligation.

BurningEnlightenment:
> That said, I have some comments, which I will place in the code. It would be nice if you could have a look at them, but you are, of course, under no obligation.

I appreciate your feedback and will work towards an implementation we are both comfortable with. That said this one of my CMake templates with some necessary modifications and it didn't get too much love, because I was in a hurry (i.e. no NEON support etc.). Please note that (as stated in the copyright notice) I'm not the author of the `TargetArch.cmake` script which I haven't touched unitl now.

I think I'll answer the rest of your questions in place.

BurningEnlightenment:
@aklomp can you estimate when you can answer my questions? I don't want to pressure you, but I'm currently planning my weekend and it would be nice if I can take this into account.

aklomp:
No problem. I don't think I'll be doing much on the project today as I have other plans this evening, and the same is mostly true for tomorrow. (I might have some time to work on this project, but no guarantees.) On Sunday, I actually have to be in the office for a server migration, so I think it's a safe bet to say that I won't be doing much over the weekend. Also, I planned to "comment" in the form of pushing some example code, which takes a bit more time than just typing text.

BurningEnlightenment:
# Status Update
I reworked and simplified quite some bits. Currently all target platform / instruction set detection code is disabled (no third party code anymore :tada:). I included some SIMD instruction set detection code for SSSE3 and AVX2 and an infrastructure skeleton for it located in `TargetSIMDInstructionSet.cmake` just in case you want to add those checks in the future.

I did some research regarding NEON, but I need an ARM platform in order to check the cmake behavior on that platform and the implications for cross compiling, i.e. that won't be added before my Odroid-C1+ arrives...

Did I forget something?

BurningEnlightenment:
Just added cmake projects for the test driver and the benchmark.

aklomp:
Looks good, I'll review it on short notice! Do you consider it fit to merge as-is? Don't worry about the NEON stuff, I have a Raspberry Pi 2B which has NEON32 extensions, and I can run NEON64 in an emulator. We can iterate.

Sorry for all the delay btw. Your pull request has been on my radar for the whole time, but I've also been occupied with implementing a faster algorithm (currently in Master), and working on merging OpenMP support (see Issue #9).

BurningEnlightenment:
> Do you consider it fit to merge as-is?

The scripts work well enough for the plain, SSSE3 and AVX2 use cases to be merged :shipit:. And as you said we ~~can~~ will iterate :wink:. I'll probably make (a) new pull request(s) for added OpenMP support and NEON stuff after I passed my exams.

> implementing a faster algorithm (currently in Master), and working on merging OpenMP support

Awesome work! :+1:

aklomp:
Just checked out some of your CMake stuff. I'm definitely new at this, but I'm open to learn. I wanted to build the benchmarks with AVX2 support, so I tried this:

```sh
cmake -DBASE64_WITH_AVX2=1 .
make
cd test
cmake .
make
```

This series of commands fails for me. It looks like the generated Makefiles in `test/` want to link with a system-wide `-lbase64`, while they should actually just include `../lib/libbase64.o` as an extra object file. Except that that file is not built anymore. Instead there is a `libbase64.a` in the project root, which exports way too many global symbols. (The list of "allowed" global symbols is in `lib/exports.txt`.) If I remove the `target_link_libraries` call in `test/CMakeLists.txt` and add `../libbase64.a` as an extra source file, things break even more. Am I doing something wrong?

Other things I noticed: the `bin/base64` binary is not built by default, and `make clean` no longer fully works. It doesn't purge generated stuff like `CMakeFiles/`.

aklomp:
FWIW, I managed to compile the tests by adding the following to `test/CMakeLists.txt`:

```cmake
find_library(LIBBASE64 libbase64.a ..)
target_link_libraries(${TEST_NAME} ${LIBBASE64})
```

BurningEnlightenment:
Well, try this (after you completely cleaned the project dir):
```sh
cd /path/to/project/src
mkdir build
cd build

cmake -DBASE64_WITH_AVX2=1 ..
make
make test
```
## Explanation
`make clean` only removes the build output, but not the build files which were generated by cmake. Because of this and because generating a `Makefile` in the project root will overwrite *your* `Makefile` it is advised to make an out of source build, i.e. create a "build" directory where all the temporary build files reside (this doesn't necessarily have to be a subdirectory of the project root). In this scenario you don't do `make clean`, but `rm -rf build`. If you use a project subdirectory like `build/` I recommend you adding `build/` to your global gitignore file.

Furthermore cmake likes hierarchies, so `test/CMakeLists.txt` isn't really a separate cmake project which can be used/build on its own. Instead it is called from the root `CMakeLists.txt` providing the necessary context like where our library is located (`FindLibrary(...)` will find you your *system* base64 library...). This also means that the `cmake ...` command generates one master `Makefile` in the "build" directory which is used for everything.

One more thing to note: `make test` will not (re-)build the test executables, it'll only execute them and report whether one of them crashed, but I've included the test executables in the `all` target, so that they will get rebuild if anything (library sources, executable sources, build files) changes.
Furthermore you probably want to use either [ccmake](https://cmake.org/cmake/help/v3.3/manual/ccmake.1.html) or [cmake-gui](https://cmake.org/cmake/help/v3.3/manual/cmake-gui.1.html), because they list all available options and let you easily tweak the compiler arguments.

BurningEnlightenment:
@aklomp are you still interested in this? If yes I will invest some time updating the build files soon:tm:.

aklomp:
Sorry about the silence, I didn't intend to string you along. I am definitely still interested in `cmake` integration (because it's the de-facto cross-platform C buildtool), and I think your work is a very good starting point. I also think you tackled all the comments I made previously, so I should do my share. What's preventing me from merging it directly is mostly this:

- I'm unfamiliar with `cmake`, but if I merge this, I would have to support it from now on. This is my problem, not yours, but it means that I need to spend some time learning about it before I can commit to it. I'm a man of many projects and great laziness, so I've been putting it off.

- Related: I feel like I do not have the `cmake` knowledge to grade the quality level of your contribution. I mean, it looks fine, but maybe it's not idiomatic or whatever. I can't really get my finger behind it, which makes me hesitant to accept it. (Again, my problem.)

- I'm not sure I like `cmake` very much. This is totally irrelevant in the light of it being the de-facto cross-platform C buildtool, but it explains something about the points above. I recognize my own bias here.

Maybe this is how we can proceed:

- I should spend some time learning about `cmake`, no excuses.
- I can merge the code into a staging branch for the time being, so that it becomes part of the master project, and proceed from there.

BurningEnlightenment:
I don't like cmake either, but sadly as you said it's a de-facto standard and just works:tm: :smile: 

> * I should spend some time learning about cmake, no excuses.

:+1: 

> * I can merge the code into a staging branch for the time being, so that it becomes part of the master project, and proceed from there.

Yeah, I would like to get this into the upstream project, but I think we should wait with merging until you can maintain the build files yourself. Instead I will watch your repository and keep the cmake files in my fork in sync with the upstream repository. I suggest you add a link in the project description referring to the fork; this way I don't have to open new pull requests every time something changes and people can use the cmake tooling now, so:
* I will close this pull request and rename my `base64` to `base64-cmake` and move the cmake files to the `master` branch
* I continue to maintain the cmake files for now
* You learn cmake
* New pull request + merge of the build files into the upstream repository
* You maintain the build files

Are you :ok: with this?

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+
+endif()
+
+###################################################################
+# OpenMP
+if(BASE64_WITH_OPENMP)
+    find_package(OpenMP)
+    if (OPENMP_FOUND)
+        option(BASE64_USE_OpenMP "Utilize OpenMP to parallelize encoding and decoding." ON)
+    endif()
```
madebr, line 53: Having `BASE_WITH_OPENMP` and `BASE64_USE_OpenMP` is weird.
In my experience, it is better to fail hard when a desired option (`BASE64_WITH_OPENMP`) is not found.
```suggestion
    find_package(OpenMP REQUIRED)
```

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+#elseif (_TARGET_ARCH STREQUAL "arm64" AND BASE64_WITH_NEON64)
+
+endif()
+
+configure_file("${CMAKE_CURRENT_LIST_DIR}/cmake/config.h.in" "${CMAKE_CURRENT_BINARY_DIR}/config.h" @ONLY)
+target_include_directories(base64 PRIVATE "${CMAKE_CURRENT_BINARY_DIR}")
+
+########################################################################
+# OpenMP Settings
+if (BASE64_USE_OpenMP)
```
madebr, line 163:
```suggestion
if (BASE64_WITH_OPENMP)
```

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+if(BASE64_WITH_OPENMP)
+    find_package(OpenMP)
+    if (OPENMP_FOUND)
+        option(BASE64_USE_OpenMP "Utilize OpenMP to parallelize encoding and decoding." ON)
+    endif()
+endif()
+
+########################################################################
+# library project
+add_library(base64 STATIC
```
madebr, line 58: See https://github.com/aklomp/base64/issues/70

I think shared libraries are possible.

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+    lib/arch/neon32/codec.c
+    lib/arch/neon64/codec.c
+)
+
+if (NOT BASE64_STANDALONE_PROJECT)
+    # export inlucde path if base64 added as subdirectory
+    target_include_directories(base64 INTERFACE
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    )
+endif()
```
madebr, line 82: `BASE64_STANDALONE_PROJECT` is not defined anywhere.
And there is no harm in adding include directories whether it is standalone or not, or is it?

```suggestion
target_include_directories(base64 PUBLIC
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
)
endif()
```

madebr (commented):
@aklomp 
We just received a request for inclusion in conan-center-index of base64 at https://github.com/conan-io/conan-center-index/pull/3746.
He used this cmake build script as a basis for providing Visual Studio support.
Having such support upstream would greatly simplify things.
You can always ping me when you have some problem.
Having cmake support would also simplify github actions on all 3 major platforms.
You get also support for shared libraries, fixing #70. This cmake pr will need some extra code for supporting DLL's though.

@BurningEnlightenment 
Great cmake script! I like your arch detection.
I have some comments about the installation, but let's address these after my first comments.

aklomp:
Hi @madebr and @BurningEnlightenment,

Yes, it's high time we find a way to merge this cmake support into master. It's probably the most requested feature, and one of the oldest open issues. After putting it off for more than two years, I have no further excuses to postpone its inclusion.

I have to say that I'm still not familiar with cmake nor do I like the smell of it, which will make it less comfortable to maintain, but I'm willing to make that my problem. Same thing for the necessary changes to `README.md`- those are on me.

What's left? Apart from the comments above, I ran into these issues:

- I noticed that `feature/cmake` no longer cleanly rebases against the current master.
- I tried to build the library using the steps given in a [previous comment](https://github.com/aklomp/base64/pull/7#issuecomment-189192945), but I get a build error because `../../config.h` cannot be found. Is that just my setup?

madebr:
@aklomp 
I'm working on it now. Is it ok if I open a new pr and merge with current master instead of rebase?

aklomp:
Let's give @BurningEnlightenment some time to respond first. If it's all the same, I'd like to merge cmake support under this specific issue, because I think the original author deserves some recognition/satisfaction after waiting this long.

You can of course open a new PR for your branch, though we'd be left with two PRs implementing the same basic change. I don't see much harm in it though, and it may be a more natural way to collaborate on any issues with your specific fixes.

As for the merge strategy, my preference is to rebase on `master` and then merge, because it gives the cleanest history. The log of recent commits for this project show what I have in mind. Each issue has a set of commits implementing one atomic change, and there's a merge commit which ties that branch to master. I'd prefer to follow that approach for this patchset too, because the changes are additive in nature and should compose cleanly on top of the rest of the codebase.

madebr:
I will of course do my patches on top of his work, so unless you squash the patches no history will be lost.

madebr:
@aklomp 
I have opened https://github.com/BurningEnlightenment/base64-cmake/pull/3 with my suggestions.

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+
+########################################################################
+if (BASE64_BUILD_TESTS)
+    enable_testing()
+    add_subdirectory(test)
+endif()
+
+########################################################################
+# install target
+if (BASE64_INSTALL_TARGET)
```
MrPointer, line 178: Don't put this in a conditional, install instructions should always exist - They simply won't be executed unless the `install` target is built.

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+    install(TARGETS base64 EXPORT base64-targets
+        RUNTIME DESTINATION bin/$<CONFIG>
+        LIBRARY DESTINATION lib/$<CONFIG>
+        ARCHIVE DESTINATION lib/$<CONFIG>
+        INCLUDES DESTINATION include
```
MrPointer, line 183: Better avoid stating those dirs manually, rather, use the `GNUInstallDirs` module: `include(GNUInstallDirs)`.  
It would then look like this:
```suggestion
        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
        INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
```

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+    )
+    install(FILES include/libbase64.h DESTINATION include)
+
+    include(CMakePackageConfigHelpers)
+    write_basic_package_version_file(
+        "${CMAKE_CURRENT_BINARY_DIR}/base64-config-version.cmake"
+        VERSION ${BASE64_VERSION}
+        COMPATIBILITY ExactVersion
+    )
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/base64-config-version.cmake" DESTINATION cmake)
```
MrPointer, line 193: Use `${INSTALL_CONFIGDIR}` as the `DESTINATION`

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+
+    configure_file(cmake/base64-config.cmake
+        "${CMAKE_CURRENT_BINARY_DIR}/base64-config.cmake"
+        COPYONLY
+    )
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/base64-config.cmake" DESTINATION cmake)
+
+    install(EXPORT base64-targets DESTINATION cmake)
```
MrPointer, line 201: Use `${INSTALL_CONFIGDIR}` as the `DESTINATION`

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+        VERSION ${BASE64_VERSION}
+        COMPATIBILITY ExactVersion
+    )
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/base64-config-version.cmake" DESTINATION cmake)
+
+    configure_file(cmake/base64-config.cmake
+        "${CMAKE_CURRENT_BINARY_DIR}/base64-config.cmake"
+        COPYONLY
+    )
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/base64-config.cmake" DESTINATION cmake)
```
MrPointer, line 199: Use `${INSTALL_CONFIGDIR}` as the `DESTINATION`

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+        ARCHIVE DESTINATION lib/$<CONFIG>
+        INCLUDES DESTINATION include
+    )
+    install(FILES include/libbase64.h DESTINATION include)
+
+    include(CMakePackageConfigHelpers)
+    write_basic_package_version_file(
+        "${CMAKE_CURRENT_BINARY_DIR}/base64-config-version.cmake"
+        VERSION ${BASE64_VERSION}
+        COMPATIBILITY ExactVersion
```
MrPointer, line 191: Why not `AnyNewerVersion`?

CMakeLists.txt
```diff
@@ -0,0 +1,202 @@
...
+    )
+    install(FILES "${CMAKE_CURRENT_BINARY_DIR}/base64-config-version.cmake" DESTINATION cmake)
+
+    configure_file(cmake/base64-config.cmake
+        "${CMAKE_CURRENT_BINARY_DIR}/base64-config.cmake"
+        COPYONLY
```
MrPointer, line 197: Replace it with the following:  
```suggestion
    configure_package_config_file(cmake/base64-config.cmake
        "${CMAKE_CURRENT_BINARY_DIR}/base64-config.cmake"
         INSTALL_DESTINATION ${INSTALL_CONFIGDIR}
)
```

MrPointer:
Sorry for jumping in the middle, I went straight to the code and noticed the review comments before I even read the discussion here 😅  
Anyway, I left my comments here as well, especially regarding installation. @madebr Looks like you wanted to do so anyway, so again, sorry if I interrupted the process, I can delete the review if you'd like me to.  

Basically, CMake and Conan support are the only things keeping me from using it in a work project, so I'm glad there's work going on in this area.

CMakeLists.txt
madebr: This ⬆️ .
There is no need to emulate autotools behavior with building shared and static libraries simultaneously.
Users will use either static or shared libraries.

CMakeLists.txt
madebr: *Hiding* the `install`'s with an `if` is useful when using `base64` as a subproject.

madebr:
@MrPointer 
I'm going to open a new pr with my branch + your suggestions next year.
As punishment for not reading my comments before posting, I'm going to tag you for extra review 😉😛 
Thanks for the checking on this.

Poking @BurningEnlightenment to see whether (s)he is alive.

MrPointer:
@madebr With pleasure 😊

shamoya:
Hi all,
Is this (or any other CMake support) merged into one of the branches of the project?
I would be delighted to use it if this is the case (my project is heavily integrated with CMake).

CMakeLists.txt
BurningEnlightenment: I settled with switching to `cmake_dependent_option` and diplaying the approppriate information with cmake's feature summary.

CMakeLists.txt
BurningEnlightenment: My opinion regarding this has been changed by the advent of package managers like conan and vcpkg. It feels to me like the subproject case became esoteric enough to be ignored and I removed the guards while overhauling the install targets. Thoughts?

CMakeLists.txt
BurningEnlightenment: I included the `$<CONFIG>` part because of multi target generators like Visual Studio, but I no longer think that this matters much due to package managers and `CMakePresets.json`. Therefore I completely removed the path specification which is equivalent to your suggestion.

CMakeLists.txt
BurningEnlightenment: @aklomp has to decide this, I tentatively went for the conservative option.

BurningEnlightenment:
Thanks @MrPointer and @madebr for reviewing this; I finally got around to clean this up 😌

CMakeLists.txt
madebr: It's always useful to disable installation.
There are always people who prefer vendoring dependencies.

I have used the following approach, to hide if checks scattered around a project:
```cmake
set(base64_MAINPROJECT ON)
if(DEFINED PROJECT_SOURCE_DIR)
    set(base64_MAINPROJECT OFF)
endif()

project(bas64)

# Enable `BASE64_INSTALL` by default when base64 is the mainproject.
option(BASE64_INSTALL "Enable base64 installation" ${base64_MAINPROJECT})

# Use base64_install instead of install
function(base64_install)
    if(BASE64_INSTALL)
        install(${ARGN})
    endif()
endfunction()

# base64_install will only install the target base64 when `BASE64_INSTALL` is true.
base64_install(TARGETS base64
    ARCHIVE DESTINATION lib
    LIBRARY DESTINATION lib
    RUNTIME DESTINATION bin
)

```

CMakeLists.txt
madebr: `ExactVersion` is only useful it a library is not source backward compatible.
I believe base64 has some API stability.

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+# Regenerate headers
+
+if (BASE64_REGENERATE_TABLES)
+    # Generate tables in build folder and copy to source tree.
+    # Don't add the tables in the source tree to the outputs, to avoid `make clean` removing them.
+    add_executable(table_generator
+        lib/tables/table_generator.c
+    )
+
+    add_custom_command(OUTPUT table_dec_32bit.h
```
madebr, line 77: Add the header as output, to let cmake know that it is part of the dependency tree.
```suggestion
    add_custom_command(OUTPUT table_dec_32bit.h "${CMAKE_CURRENT_SOURCE_DIR}/lib/tables/table_dec_32bit.h"
```

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+        COMMAND table_generator > table_dec_32bit.h
+        COMMAND "${CMAKE_COMMAND}" -E copy table_dec_32bit.h "${CMAKE_CURRENT_SOURCE_DIR}/lib/tables/table_dec_32bit.h"
+        DEPENDS table_generator
+    )
+    set(Python_ADDITIONAL_VERSIONS 3)
+    find_package(PythonInterp REQUIRED)
+    add_custom_command(OUTPUT table_enc_12bit.h
```
madebr, line 84: Likewise

```suggestion
    add_custom_command(OUTPUT table_enc_12bit.h "${CMAKE_CURRENT_SOURCE_DIR}/lib/tables/table_enc_12bit.h"
```

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+    lib/arch/generic/codec.c
+
+    lib/arch/ssse3/codec.c
+    lib/arch/sse41/codec.c
+    lib/arch/sse42/codec.c
+    lib/arch/avx/codec.c
+    lib/arch/avx2/codec.c
+
+    lib/arch/neon32/codec.c
+    lib/arch/neon64/codec.c
```
madebr, line 112: You need to add generated sources explicitly to a target.
Otherwise, the `add_custom_command` will never fire.
```suggestion
    lib/arch/neon64/codec.c

    # Add generated headers explicitly to target, to insert them in the dependency tree
    lib/tables/table_dec_32bit.h
    lib/tables/table_enc_12bit.h
```

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+    target_compile_definitions(base64 
+        PRIVATE
+            BASE64_EXPORTS
+    )
+else()
+    target_compile_definitions(base64
+        PUBLIC
+            BASE64_STATIC_DEFINE
+    )
+endif()
```
madebr, line 149: `DEFINE_SYMBOL`: https://cmake.org/cmake/help/latest/prop_tgt/DEFINE_SYMBOL.html
-> it's only defined when building as a shared library.
```suggestion
    SOVERSION ${PROJECT_VERSION_MAJOR}
    DEFINE_SYMBOL BASE64_EXPORTS
)

#generate_export_header(base64)
# the following definitions and those in libbase64.h have been
# kept forward compatible in case we ever switch to generate_export_header
if (BUILD_SHARED_LIBS)
    set_target_properties(base64 PROPERTIES
        C_VISIBILITY_PRESET hidden
    )
else()
    target_compile_definitions(base64
        PUBLIC
            BASE64_STATIC_DEFINE
    )
endif()
```

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+)
+
+install(FILES
+        "${CMAKE_CURRENT_BINARY_DIR}/base64-config.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/base64-config-version.cmake"
+    DESTINATION "${CMAKE_INSTALL_DATADIR}/base64"
+)
+
+install(EXPORT base64-targets
+    NAMESPACE aklomp::
```
madebr, line 258: I would just use `base64` as namespace.
```suggestion
    NAMESPACE base64::
```

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+########################################################################
+# cmake install
+install(DIRECTORY include/ TYPE INCLUDE)
+install(TARGETS base64 EXPORT base64-targets)
+
+include(CMakePackageConfigHelpers)
+configure_package_config_file(cmake/base64-config.cmake.in
+    "${CMAKE_CURRENT_BINARY_DIR}/base64-config.cmake"
+
+    INSTALL_DESTINATION "${CMAKE_INSTALL_DATADIR}/base64"
```
madebr, line 243: cmake scripts are commonly installed to `${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}`

BurningEnlightenment: Have people finally settled on that convention? The [`install(EXPORT)`] and [`configure_package_config_file()`] docs don't explicitly recommend a destination, but provide examples with the following conflicting patterns:
- `$LIBDIR/${PROJECT_NAME}`
- `$LIBDIR/${PROJECT_NAME}/cmake`

Then of course there is the plethora other of options supported by [`find_package()`]. I usually opt for `$DATADIR/${PROJECT_NAME}` because that's the default for [`vcpkg_cmake_config_fixup()`] which in turn makes my [`portfile.cmake`] a little bit leaner.

However, I have no strong feelings one way or the other.

[`install(EXPORT)`]: https://cmake.org/cmake/help/latest/command/install.html?highlight=DESTINATION#export
[`configure_package_config_file()`]: https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html?highlight=INSTALL_DESTINATION#example-generating-package-files
[`find_package()`]: https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure
[`vcpkg_cmake_config_fixup()`]: https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/ports/vcpkg-cmake-config/vcpkg_cmake_config_fixup.md
[`portfile.cmake`]: https://github.com/deeplex-net/vcpkg-registry/blob/master/ports/base64/portfile.cmake

madebr: It's just a matter of preference, I guess. Both are supported by [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure)
On my Fedora Linux systems, all cmake scripts, except one, are installed in `/usr/lib64/cmake`.


You can always make it configurable, and use the variable everywhere.
Then everybody is happy.
e.g.
```cmake
set(BASE64_INSTALL_CMAKEDIR "${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}" CACHE PATH "Where to install cmake scripts of base64")
```

BurningEnlightenment: > It's just a matter of preference, I guess. [...]
> On my Fedora Linux systems, all cmake scripts, except one, are installed in `/usr/lib64/cmake`.

That's good enough for me and since there are no inherent technical drawbacks I'll go with your first suggestion; no need for unicorn solutions here.

> You can always make it configurable, and use the variable everywhere.

I'd like to avoid that if possible as the saying goes "All problems in computer science can be solved by another level of indirection -- except for the problem of too many layers of indirection." 😉

BurningEnlightenment:
Should we add some common [cmake presets] by the way?

[cmake presets]: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+    lib/arch/avx2/codec.c
+
+    lib/arch/neon32/codec.c
+    lib/arch/neon64/codec.c
+)
+
+target_include_directories(base64
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        $<INSTALL_INTERFACE:include>
```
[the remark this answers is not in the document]
BurningEnlightenment: The [cmake package docs] recommend not using absolute paths, because it keeps the package relocatable.

[cmake package docs]: https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages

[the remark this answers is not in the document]
madebr: Yes,  `CMAKE_INSTALL_FULL_INCLUDEDIR` contains the full path.
See the documentation of [`GnuInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables):
> CMAKE_INSTALL_<dir>
>    Destination for files of a given type. This value may be passed to the DESTINATION options of install() commands for the corresponding file type. It should typically be a path relative to the installation prefix so that it can be converted to an absolute path in a relocatable way (see CMAKE_INSTALL_FULL_<dir>). However, an absolute path is also allowed.

Technically, it can be an absolute path, but if you're doing that, then you're screwing things up anyways.

dotnwat:
Hi everyone.

What's the plan to get cmake support merged? It looks like the latest feedback has been taken into account, and having some support would be nice. It will be easier to iterate on cmake support as the community starts to use it.

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+    target_compile_definitions(base64
+        PUBLIC
+            BASE64_STATIC_DEFINE
+    )
+endif()
+
+target_compile_options(base64 PRIVATE
+  $<$<C_COMPILER_ID:MSVC>:
+    /W4
+    /WX # All warnings as error
```
madebr, line 154: To my own projects, I usually add a `PROJECT_WERROR` option to enable/disable warnings as errors (enabled by default).
That way, users have the possibility to ignore warnings.
This is useful for new compiler releases that introduce new warnings.
```cmake
option(BASE64_WERROR "Treat warnings as error" ON)
target_compile_options(base64 PRIVATE
    $<$<BOOL:${BASE64_WERROR}>:$<IF:$<C_COMPILER_ID:MSVC>,/WX,-Werror>>
)
```

Currently, only MSVC has warnings as error behavior.

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+
+    if (BASE64_WITH_NEON32)
+        set_source_files_properties("lib/arch/neon32/codec.c" PROPERTIES
+            COMPILE_FLAGS "${BASE64_NEON32_CFLAGS} "
+        )
+    endif()
+
+#elseif (_TARGET_ARCH STREQUAL "arm64" AND BASE64_WITH_NEON64)
+
+endif()
```
madebr, line 215: Perhaps add a `else()` with a warning about an unknown arch?
```cmake
else()
    message(WARNING "Unknown architecture: ${_TARGET_ARCH}")
endif()
```

CMakeLists.txt
madebr: Can you take another look at this suggestion?

include/libbase64.h
```diff
@@ -3,6 +3,39 @@
...
 #include <stddef.h>	/* size_t */
 
+
+#if defined(_WIN32) || defined(__CYGWIN__)
+#define BASE64_SYMBOL_IMPORT __declspec(dllimport)
+#define BASE64_SYMBOL_EXPORT __declspec(dllexport)
+#define BASE64_SYMBOL_PRIVATE
+
+#elif __GNUC__ >= 4
+#define BASE64_SYMBOL_IMPORT   __attribute__ ((visibility ("default")))
```
madebr, line 13: I think import symbols don't need this.
```suggestion
#define BASE64_SYMBOL_IMPORT
```

CMakeLists.txt
aklomp: Currently we [use](https://github.com/aklomp/base64/releases) a kind of semantic versioning scheme for the releases. It's conceivable that I overhaul the API for a `1.x.x` or `2.x.x` release, because there are parts that I'm not happy with. Is it possible to tell `cmake` that API compatibility is only guaranteed for the minor and patch versions?

CMakeLists.txt
aklomp: Tough call. On one hand `base64` makes sense as a namespace, but on the other hand it seems kind of invasive to claim the generic term like that.

So then use `aklomp` as the vendor string? Hm, would work, except it's my name and not a brand. I don't publish any other libraries under the `aklomp` "banner". It would be weird to require people to include it in their `CMakeLists` or how it's called.

I think `libbase64` might be a compromise?

aklomp:
@dotnwat Merging `cmake` support is with me as the maintainer of this project :crying_cat_face:

As I've written earlier in this thread, I feel conflicted about merging this PR because I have hardly any experience with `cmake`, I don't really understand the changes being introduced, and I'm not confident I can maintain them. The obvious solution to that problem was for me to start learning `cmake` so I can converse on equal footing. But I still haven't gotten round to it (made it enough of a priority), and I'm sorry about that.

(It's absolutely been on my ongoing mental to-do list for *years* and causing me quite some anguish as one of those points that *you really need to do* one day. I didn't just walk away from this. But that's not doing anyone in this thread any good.)

So at a given moment I said I was willing to merge any `cmake` config that demonstrably works. That offer still stands. So why am I not merging this PR now?

- There are several open questions in this thread (mainly by @madebr) which I don't feel qualified to answer.
- I can't get `cmake` to actually compile the library (more below). That makes me back off from merging.
- The PR does not include an update to the `README` to explain how to compile the code with `cmake`. OK, that's minor and can be fixed later.

So that's basically why it's in limbo. As ashamed as I now am by all of this, I am seriously contemplating just merging everything sight-unseen and sorting out any breakage in later PRs. I tend to agree with you that any support is better than no support, and that once it's mainlined, it can be iterated upon.

Back to the build not working for me. I'm trying to build the code with this set of commands:

```sh
cmake -B build
cmake --build build
```

But I get this build error:

```sh
[  5%] Building C object CMakeFiles/base64.dir/lib/lib.c.o
[ 11%] Building C object CMakeFiles/base64.dir/lib/codec_choose.c.o
[ 17%] Building C object CMakeFiles/base64.dir/lib/arch/generic/codec.c.o
lib/arch/generic/codec.c:8:10: fatal error: ../../config.h: No such file or directory
    8 | #include "../../config.h"
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
gmake[2]: *** [CMakeFiles/base64.dir/build.make:104: CMakeFiles/base64.dir/lib/arch/generic/codec.c.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:102: CMakeFiles/base64.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
```

Any clue what I'm doing wrong? What do I need to do to demonstrate for myself that this patchset is good?

aklomp:
One large problem with this PR is that everything is compiled without optimization. The result is that the code is incredibly slow. All the tricks that we do to get good performance (SIMD, fancy algorithms, aggressive inlining) is all for nothing: `plain` is the fastest codec. Here's part of the benchmark compiled with stock `cmake`:

```
Testing with buffer size 100 KB, fastest of 10 * 100
AVX2    encode  146.29 MB/sec
AVX2    decode  137.06 MB/sec
plain   encode  550.22 MB/sec
plain   decode  476.14 MB/sec
SSSE3   encode  135.73 MB/sec
SSSE3   decode  146.43 MB/sec
SSE41   encode  152.59 MB/sec
SSE41   decode  154.20 MB/sec
SSE42   encode  152.85 MB/sec
SSE42   decode  154.21 MB/sec
AVX     encode  158.27 MB/sec
AVX     decode  158.16 MB/sec
```

Here's what it should look like:

```
Testing with buffer size 100 KB, fastest of 10 * 100
AVX2    encode  13745.55 MB/sec
AVX2    decode  14467.66 MB/sec
plain   encode  3475.35 MB/sec
plain   decode  3259.66 MB/sec
SSSE3   encode  7606.59 MB/sec
SSSE3   decode  8212.31 MB/sec
SSE41   encode  7603.66 MB/sec
SSE41   decode  8401.14 MB/sec
SSE42   encode  7600.30 MB/sec
SSE42   decode  8211.38 MB/sec
AVX     encode  7996.93 MB/sec
AVX     decode  9094.13 MB/sec
```

This second run is with a benchmark tool made by the old `Makefile`, which [optimizes at `-O3`](https://github.com/aklomp/base64/blob/master/Makefile#L1) by default.

This is clearly not something we want. Can `cmake` be instructed to always optimize to `-O3` too?

CMakeLists.txt
aklomp: A warning definitely makes sense.

CMakeLists.txt
```diff
@@ -0,0 +1,263 @@
...
+
+target_compile_options(base64 PRIVATE
+  $<$<C_COMPILER_ID:MSVC>:
+    /W4
+    /WX # All warnings as error
+    /we4013 # Error warning C4013: 'function' undefined; assuming extern returning int
+    /we4700 # Error warning C4700: uninitialized local variable
+    /we4715 # not all control paths return a value
+    /we4003 # not enough actual parameters for macro
+    /wd4456 # disable warning C4456: declaration of 'xxx' hides previous local declaration
```
aklomp, line 159: If all of these are legitimate issues with the code, I'd prefer to have them fixed rather than ignore the warning.

cmake/Modules/TargetSIMDInstructionSet.cmake
```diff
@@ -0,0 +1,34 @@
...
+        set(COMPILE_FLAGS_AVX "-mavx")
+        set(COMPILE_FLAGS_AVX2 "-mavx2")
+
+        #arm
+        set(COMPILE_FLAGS_NEON32 "-mfpu=neon")
+    elseif(MSVC)
+        set(COMPILE_FLAGS_SSSE3 " ")
+        set(COMPILE_FLAGS_SSE41 " ")
+        set(COMPILE_FLAGS_SSE42 " ")
+        set(COMPILE_FLAGS_AVX "/arch:AVX2")
```
aklomp, line 31: This looks like a typo. The `AVX` codec actually falls back to the `ssse3` codec. It should not be compiled as `AVX2`, because the target machine might not support that SIMD level.

CMakeLists.txt
BurningEnlightenment: Yes, that would be [`SameMajorVersion`](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html?highlight=SameMajorVersion#generating-a-package-version-file).
It is worth noting that downstream users always have the option to accept any version by calling [`find_package()`](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-version-selection) without a version. The downstream user may also insist on an exact version match.

CMakeLists.txt
BurningEnlightenment: AFAIK there are two widely used conventions: Using a vendor/organization string or using the cmake project name (which in our case would be `base64`). I dislike `base64::base64`, because it is unlikely to prevent target name collisions. 
However, in the end this is just bikeshedding, so I will change this to whatever you choose as soon as you choose.

CMakeLists.txt
BurningEnlightenment: I concur.
What do you think about the warnings which are explicitly marked as errors? I don't quite remember adding them 😅

CMakeLists.txt
BurningEnlightenment: Does it though? Obviously there won't be SIMD acceleration, because SIMD instruction sets are architecture specific and one chose to compile for an alien one. Or is this more about not providing official support/ not regularly testing the pure C codec on that architecture?

include/libbase64.h
BurningEnlightenment: I'm not an expert w.r.t. gcc symbol visibility; I looked at the [`GenerateExportHeader`](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) output and noticed that they do define the import macro with said attribute (s. [`GenereateExportHeader.cmake:310`](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/GenerateExportHeader.cmake#L310)). However, I don't know the rationale.

cmake/Modules/TargetSIMDInstructionSet.cmake
```diff
@@ -0,0 +1,34 @@
...
+        set(COMPILE_FLAGS_SSE42 "-msse4.2")
+        set(COMPILE_FLAGS_AVX "-mavx")
+        set(COMPILE_FLAGS_AVX2 "-mavx2")
+
+        #arm
+        set(COMPILE_FLAGS_NEON32 "-mfpu=neon")
+    elseif(MSVC)
+        set(COMPILE_FLAGS_SSSE3 " ")
+        set(COMPILE_FLAGS_SSE41 " ")
+        set(COMPILE_FLAGS_SSE42 " ")
```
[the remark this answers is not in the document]
BurningEnlightenment: To the contrary the SIMD intrinsics up to and including `SSE42` are always available on the MSVC platform and cannot be disabled. This isn't an issue for MSVC because they don't utilize them for auto vectorization (unlike gcc).

cmake/Modules/TargetSIMDInstructionSet.cmake
BurningEnlightenment: Good catch!

CMakeLists.txt
BurningEnlightenment: There is currently one violation in the windows [`__cpuid`](https://github.com/aklomp/base64/blob/master/lib/codec_choose.c#L23) macro definition and its [usage in `codec_choose_x86()`](https://github.com/aklomp/base64/blob/master/lib/codec_choose.c#L174). Sadly one can only disable the warning as a whole and not just it being treated as an error.

BurningEnlightenment:
> One large problem with this PR is that everything is compiled without optimization.

CMake has four different build types (`Debug`, `MinSizeRel`, `RelWithDebInfo` and `Release`) of which it defaults to `Debug` which as you noticed is an unoptimized build. The other build types have the usual optimization flags set which can also be overriden by the user. You can explicitly specify a build type like so:
```
cmake -B build -DCMAKE_BUILD_TYPE=Release
```
I recommend using the more graphical [`ccmake`](https://cmake.org/cmake/help/latest/manual/ccmake.1.html) rather than `cmake` if you want to inspect/ fiddle with the compiler flags.

aklomp:
@BurningEnlightenment I still think it's reasonable to set `-O3` (and the equivalent for MSVC) on all builds, because the library is pretty crap without them. With the default build, you'd be better off using a plain C library than using any of the SIMD codecs. The Makefile also sets `-O3` unconditionally, so there's precedence.

I'd like to add it unless it violates a core `cmake` principal and horrifies the gods.

dotnwat:
If there are optimizations that are only available at -O3 then it is reasonable to hard code that for all compilers. I see on my system that -O3 is the default for GCC in cmake (with `CMAKE_BUILD_TYPE=Release`), but it doesn't look like clang is specifying anything specifically.

You can always use -O3 for everything and let exceptions be handled as they arise.

CMakeLists.txt
madebr: Another comment, shouldn't ´TARGET_ARCH´ be ´HOST_ARCH´?
A target architecture only makes sense for a code generator (=compiler).
Here, we're just converting some sources to binaries (on a build machine to some host machine).

include/libbase64.h
madebr: My interpretation is that adding `visibility("default")`, you're instrumenting the compiler to export a symbol.
So adding this as a consumer will cause you to re-export the symbol.

CMakeLists.txt
```diff
@@ -0,0 +1,264 @@
...
+
+include(GNUInstallDirs)
+include(CMakeDependentOption)
+include(FeatureSummary)
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
+
+#######################################################################
+# platform detection
+include(TargetArch)
+detect_target_architecture(_TARGET_ARCH)
```
madebr, line 25: CMake uses cmake toolchains for specifying the target architectures.
See [CMAKE_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html).
Perhaps we should first try to parse `CMAKE_SYSTEM_PROCESSOR`, and then try to run the compiler?
Using that approach, users can override the architecture in the case of a mis-detection.

CMakeLists.txt
```diff
@@ -0,0 +1,264 @@
...
+# the following definitions and those in libbase64.h have been
+# kept forward compatible in case we ever switch to generate_export_header
+if (BUILD_SHARED_LIBS)
+    set_target_properties(base64 PROPERTIES
+        C_VISIBILITY_PRESET hidden
+    )
+else()
+    target_compile_definitions(base64
+        PUBLIC
+            BASE64_STATIC_DEFINE
```
madebr, line 148: I find the ´_DEFINE´ suffix weird: it's a definition anyways.
How about using simply `BASE64_STATIC`?

madebr:
> @BurningEnlightenment I still think it's reasonable to set `-O3` (and the equivalent for MSVC) on all builds, because the library is pretty crap without them. With the default build, you'd be better off using a plain C library than using any of the SIMD codecs. The Makefile also sets `-O3` unconditionally, so there's precedence.

> I'd like to add it unless it violates a core `cmake` principal and horrifies the gods.

I think we can keep the cmake gods happy by only doing "-O3" when building a Release or RelWithDebInfo config.
Then, we can keep the lower optimization level for Debug builds, so debugging remains easier.

Something as follows: (untested)
```cmake
if(NOT CMAKE_C_COMPILER_ID STREQUAL "MSVC")
    string(REGEX REPLACE "([/-])O[0-2]" "" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
    string(REGEX REPLACE "([/-])O[0-2]" "" CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO}")
    set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE } -O3")
    set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -O3")
endif()
```

CMakeLists.txt
dotnwat: Or you could effectively transliterate the makefile into cmake (ie require users to set options). I'm all about making this as feature rich as possible, but it doesn't seem like it makes sense to wait on the perfect cmake build.

htot:
I am writing a Yocto project recipe for this library and tools and are basing it on the cmake. Because the high level of standardization saves me work. It would be nice to base the recipe of this repo instead of @BurningEnlightenment fork. Can we get it in and then optimize later?

> I still think it's reasonable to set -O3 (and the equivalent for MSVC) on all builds, because the library is pretty crap without them. 

The purpose of the debug build is that you can debug in a sensible way, so optimizations _need_ to be disabled. If you really need to debug optimized code (and willing to suffer the big pain) there is the RelWithDebInfo flag.

Has anybody seen a significant speedup from -O2 to -O3?

aklomp:
Thanks everyone for your patience, the @BurningEnlightenment `cmake` branch has finally been merged! I recognize that @htot has a point in asking for the branch to be merged first and incremental improvements to be made later, even though it goes against my perfectionist tendencies :)

Anyway, sorry this took so long. Let's iterate on this in new issues.

BurningEnlightenment:
I will leave the fork up for a month or two in order to allow people to migrate their CI/ package manager scripts gracefully. 

I'd also like to thank everyone involved for the reviews, feedback and contributions 🙂

dotnwat:
Happy to see this merged. And great timing; I was just about to put up a PR to modify the relative include paths, as we are using Meson to integrate into our build system and it requires out-of-tree builds.

PR #11, opened 2016-06-27
Fix NEON64 backed build

yazevnul:


aklomp:
Looks good, thanks! The single question I have is why you bothered with the loop variable, when this library (proudly) uses C99 features everywhere. After that change, there's still C99 left in `neon.c`, such as a C99-style [inline variable definition](https://github.com/yazevnul/base64/blob/fdf54e38d00867649b0201f201652d03b2b101ec/lib/dec/neon.c#L43).

yazevnul:
Yep, that is strange, but compiler gave an error for the `for` loop :) I could put `i` back into the loop body if you want.

aklomp:
Is this with a mainstream compiler? If so, you can leave it, but I would find it strange that *anything* that compiles NEON64 does not support C99 syntax. And `-std=c99` is set inside the `Makefile`, so we don't need to worry about a missing compiler flag.

yazevnul:
It was GCC 4.9.2.

aklomp:
Why is it not compiling in C99 mode?

yazevnul:
I am not the one who set compiler flags. Actually, I am not building with Makefile, I use in-house build system where build flags are mostly the same for all projects.

aklomp:
Hm, then maybe it would be better to maintain this in a separate branch as a site-specific modification. If you put the `i` back into the loop body, I'll merge the rest.

yazevnul:
fixed

PR #12, opened 2016-06-29
fix MSVC build

yazevnul:
see https://msdn.microsoft.com/en-us/library/a3140177.aspx to read about MSVC intrinsics

### DIFF
--- a/lib/codec_avx2.c
+++ b/lib/codec_avx2.c
@@ -1,5 +1,6 @@
 #include <stdint.h>
 #include <stddef.h>
+#include <stdlib.h>
 
 #include "../include/libbase64.h"
 #include "codecs.h"
--- a/lib/codec_neon32.c
+++ b/lib/codec_neon32.c
@@ -4,6 +4,7 @@
 
 #include <stdint.h>
 #include <stddef.h>
+#include <stdlib.h>
 #ifdef __ARM_NEON__
 #include <arm_neon.h>
 #endif
--- a/lib/codec_neon64.c
+++ b/lib/codec_neon64.c
@@ -4,6 +4,7 @@
 
 #include <stdint.h>
 #include <stddef.h>
+#include <stdlib.h>
 #ifdef __ARM_NEON__
 #include <arm_neon.h>
 #endif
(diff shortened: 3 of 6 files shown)


aklomp:
Looks good, but instead of checking for `__WIN32__` and friends, couldn't you explicitly check for MSVC? Maybe GCC on Win32 also defines `__WIN32__`.

BurningEnlightenment:
> Looks good, but instead of checking for `__WIN32__` and friends, couldn't you explicitly check for MSVC? Maybe GCC on Win32 also defines `__WIN32__`.

You're right; this is a MSVC specific fix and enabling it if `_WIN32` or `__WIN32__` are defined _will_ break gcc and clang builds on windows. An appropriate preprocessor check would be:
```c++
#if defined(_MSC_VER)
```

yazevnul:
Yep, I'll fix this, was a little busy

BurningEnlightenment:
Errr, I'm very curious how you came across this... those intrinsics are only used in `(enc|dec)/uint(32|64).c` and because Visual Studio doesn't define `__WORDSIZE` at all, none of those files should have been included. I just noticed that, because I compiled the test executable using my cmake build files and visual studio _without_ applying your proposed fix. To my very amusement it compiled just fine and the tests passed without any errors :smile:.
However, I'll merge @yazevnul fix into my base64-cmake repository and correct the preprocessor checks over there.

aklomp:
Just pushed a fixed version to Master. I amended the original commit, so authorship is properly attributed.

BurningEnlightenment:
@aklomp you haven't fixed the `#if __WORDSIZE == *` in `codec_plain.c` :disappointed:, whithout this MSVC won't include the `(enc|dec)/uint(32|64).c` files, see BurningEnlightenment/base64-cmake@d405e70fe0ab87ada718e209b742b8d634a72939

aklomp:
Good catch, I totally overlooked that. I just rebased that commit and the "MSVC OpenMP incompetence" (seriously, MSVC, what the hell) onto master and pushed it. Congratulations on your first in-tree commits ;)

I decided to rebase instead of merge, because your branch is built on top of the Cmake layer, and I'm not quite ready to tackle that just yet.

PR #19, opened 2016-10-15
Fix invalid input handling

mayeut:
Decoder can succeed with invalid input.
This fixes this issue and add tests for this case.

### DIFF
--- a/lib/dec/tail.c
+++ b/lib/dec/tail.c
@@ -3,7 +3,7 @@
 			break;
 		}
 		if ((q = base64_table_dec[*c++]) >= 254) {
-			st.eof = 1;
+			st.eof = BASE64_EOF;
 			// Treat character '=' as invalid for byte 0:
 			break;
 		}
@@ -15,7 +15,7 @@
 			break;
 		}
 		if ((q = base64_table_dec[*c++]) >= 254) {
-			st.eof = 1;
+			st.eof = BASE64_EOF;
 			// Treat character '=' as invalid for byte 1:
 			break;
 		}
@@ -29,11 +29,24 @@
 			break;
 		}
 		if ((q = base64_table_dec[*c++]) >= 254) {
-			st.eof = 1;
-			// When q == 254, the input char is '='. Return 1 and EOF.
-			// Technically, should check if next byte is also '=', but never mind.
-			// When q == 255, the input char is invalid. Return 0 and EOF.
-			ret = (q == 254) ? 1 : 0;
+			// When q == 254, the input char is '='.
+			// Check if next byte is also '=':
+			if (q == 254) {
+				if (srclen-- != 0) {
+					// EOF:
+					st.eof = BASE64_EOF;
+					ret = (base64_table_dec[*c++] == 254) ? 1 : 0;
+					break;
+				}
+				else {
+					// Almost EOF
+					st.eof = BASE64_AEOF;
+					ret = 1;
+					break;
+				}
+			}
+			// If we get here, there was an error:
+			ret = 0;
 			break;
 		}
 		*o++ = st.carry | (q >> 2);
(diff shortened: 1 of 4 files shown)


lib/dec/tail.c
```diff
@@ -30,10 +30,10 @@
 		}
 		if ((q = base64_table_dec[*c++]) >= 254) {
 			st.eof = 1;
-			// When q == 254, the input char is '='. Return 1 and EOF.
-			// Technically, should check if next byte is also '=', but never mind.
+			// When q == 254, the input char is '='.
+			// Check if next byte is also '=', Return 1 and EOF.
 			// When q == 255, the input char is invalid. Return 0 and EOF.
-			ret = (q == 254) ? 1 : 0;
+			ret = ((q == 254) && (base64_table_dec[*c++] == 254)) ? 1 : 0;
```
aklomp: This is incorrect; the loop may only look at one byte at a time. It cannot assume that the next byte exists.

One way of checking if there is a byte available is whether `srclen` is larger than 0, but even then you need to handle the case where there is no next byte (yet).

mayeut: Ok, done. I pushed corrections.

lib/dec/tail.c
```diff
@@ -29,11 +29,24 @@
...
 		if ((q = base64_table_dec[*c++]) >= 254) {
-			st.eof = 1;
-			// When q == 254, the input char is '='. Return 1 and EOF.
-			// Technically, should check if next byte is also '=', but never mind.
-			// When q == 255, the input char is invalid. Return 0 and EOF.
-			ret = (q == 254) ? 1 : 0;
+			// When q == 254, the input char is '='.
+			// Check if next byte is also '=':
+			if (q == 254) {
+				if (srclen-- != 0) {
```
aklomp: `srclen` is just the number of bytes being currently fed into the decoder. It is not the total length of the encoded string, since we don't know that per se. (This is a stream decoder.) So the use of `srclen` here is erroneous.

mayeut: Nope, I checked that. That's why there's the new flag `aeof` in case of streaming mode.

include/libbase64.h
```diff
@@ -18,6 +18,7 @@ extern "C" {
 
 struct base64_state {
 	int eof;
+	int aeof;
```
mayeut: yes, could be done.
I guess that structure is for internal use anyway. Why isn't it opaque structure ? To avoid `malloc`/ `free` inside the lib ?

aklomp: Indeed. The caller is free to allocate the structure as he pleases, including on the stack. I wanted to avoid all dynamic memory troubles.

lib/dec/tail.c
```diff
@@ -29,11 +29,24 @@
...
+					st.eof = 1;
+					ret = (base64_table_dec[*c++] == 254) ? 1 : 0;
+					break;
+				}
+				else {
+					// Almost EOF
+					st.aeof = 1;
+					ret = 1;
+					break;
+				}
```
aklomp: This is actually not the whole picture. If `st.eof == 1`, then the next byte MUST be `=` or it's an error.

aklomp:
```c
// If we saw the first EOF byte, this one *must* be `=`
if (st.eof == 1 && q != 254) {
        ret = 0;
        break;
}

if (q == 254) {
        st.eof++;
        break;
}

// If we get here, there was an error:
ret = 0;
break;
```

lib/dec/tail.c
mayeut: head.c was not pushed...

mayeut:
Everything has been pushed now... should be better.

aklomp:
Thanks, merged! I had perhaps some minor stylistic comments, but they were not important. Thanks especially for writing testcases, those contributions are rare in open source :)

PR #20, opened 2016-10-16
Add build support for MacOS

mayeut:
Modified Makefile to use -r instead of --relocatable so that it can be
used with MacOS linker as well.
Modified test/Makefile not to link with librt under MacOS.
Modified test/benchmark.c to use MACH time functions under MacOS.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -3,6 +3,14 @@ ifdef OPENMP
   CFLAGS += -fopenmp
 endif
 
+TARGET := $(shell $(CC) -dumpmachine)
+ifneq (, $(findstring darwin, $(TARGET)))
+  BENCH_LDFLAGS=
+else
+  # default to linux, -lrt needed
+  BENCH_LDFLAGS=-lrt
+endif
+
 .PHONY: clean test
 
 test: clean test_base64 benchmark
@@ -13,7 +21,7 @@ test_base64: test_base64.c codec_supported.o ../lib/libbase64.o
 	$(CC) $(CFLAGS) -o $@ $^
 
 benchmark: benchmark.c codec_supported.o ../lib/libbase64.o
-	$(CC) $(CFLAGS) -o $@ $^ -lrt
+	$(CC) $(CFLAGS) -o $@ $^ $(BENCH_LDFLAGS)
 
 ../%:
 	make -C .. $*
--- a/test/benchmark.c
+++ b/test/benchmark.c
@@ -9,6 +9,10 @@
 #include <stdio.h>
 #include <time.h>
 
+#ifdef __MACH__
+#include <mach/mach_time.h>
+#endif
+
 #include "../include/libbase64.h"
 #include "codec_supported.h"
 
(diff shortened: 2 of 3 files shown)


aklomp:
Thanks for the contribution! I'm definitely open to adding build support for MacOS (and a little bit surprised that it didn't compile out of the box.) However, I have some comments, which I'll attach to the code.

test/Makefile
```diff
@@ -3,6 +3,14 @@ ifdef OPENMP
   CFLAGS += -fopenmp
 endif
 
+TARGET := $(shell $(CC) -dumpmachine)
+ifneq (, $(findstring darwin, $(TARGET)))
+BENCH_LDFLAGS=
+else
+# default to linux, -lrt needed
+BENCH_LDFLAGS=-lrt
+endif
```
aklomp: Whitespace error: trailing space.

mayeut: Will remove the trailing space after `endif` and add 2 spaces for indentation as is done for in the OPENMP block

test/Makefile
```diff
@@ -3,6 +3,14 @@ ifdef OPENMP
   CFLAGS += -fopenmp
 endif
 
+TARGET := $(shell $(CC) -dumpmachine)
```
aklomp: Ah, I just remembered that the Windows builds are created using CMake, so my first point is moot.

test/benchmark.c
```diff
@@ -72,17 +76,43 @@ get_random_data (struct buffers *b, char **errmsg)
...
+timediff_sec (base64_timespec *start, base64_timespec *end)
+{
+	uint64_t diff = *end - *start;
+	mach_timebase_info_data_t tb = { 0, 0 };
+	mach_timebase_info(&tb);
+
+	return (float)((diff * tb.numer) / tb.denom) / 1e9f;
+}
+#else
+typedef struct timespec base64_timespec;
```
aklomp: Nice use of `typedef`. One of the rare times it's used correctly.

test/Makefile
aklomp: Thanks.

test/Makefile
aklomp: O_o

Ok, so I guess we're stuck with `-lrt`. But what about using `uname -s` as a slightly more portable platform check?

mayeut:
Regarding compilation out of the box on MacOS, `objcopy` is not part of the set of tools provided by Apple. One has to install binutils to get that (with `brew` for example). For some reason, once I used `OBJCOPY=gobjcopy` (at first, I just commented the line in the Makefile), the linker would complain with some message about architecture subtype. I don't know if it's a bug from `gobjcopy`, the linker or a misuse of one of them on this platform. If anyone gets that message, defining `CFLAGS="-Wl,-allow_sub_type_mismatches"` gets rid of this error (but introduces warnings about unused linker flag when doing only compilation).

test/Makefile
mayeut: `uname -s` is not available on windows & tells about the host system. If you're cross-compiling on a linux box & targeting mac, it won't work.
I guess the `cmake` PR makes more sense for true multiplatform builds. Here, I just take care of building properly on MacOS. If for some reason, the shell command fails, then it will fallback to what used to be. Only if it succeeds & you're targeting macOS/iOS will it have any effects.

aklomp: Good point on the cross compiling. OK, merged it, thanks!

PR #18, opened 2016-10-15
Add SSE4.2 code paths

mayeut:
SSE4.2 has been added for decoding.

Speed-up on `Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz` using `Apple LLVM version 8.0.0 (clang-800.0.38)`

SSE4.2 decoding: +29% compared to SSSE3 in ab7a48bcbc8066e2e200836dd9f40a2b44dda35b

Full results (this builds on top of #17):
```
Filling buffer with 10.0 MB of random data...
Testing with buffer size 10 MB, fastest of 100 * 1
AVX2	encode	7876.40 MB/sec
AVX2    decode  7047.91 MB/sec
plain   encode  1499.65 MB/sec
plain   decode  1557.21 MB/sec
SSSE3	encode	5612.09 MB/sec
SSSE3	decode	3883.52 MB/sec
SSE42	encode	5600.26 MB/sec
SSE42	decode	5089.87 MB/sec
Testing with buffer size 1 MB, fastest of 100 * 10
AVX2    encode  8970.38 MB/sec
AVX2    decode  7093.07 MB/sec
plain   encode  1501.03 MB/sec
plain   decode  1562.29 MB/sec
SSSE3	encode	5681.83 MB/sec
SSSE3	decode	3902.85 MB/sec
SSE42	encode	5681.28 MB/sec
SSE42	decode	5108.25 MB/sec
Testing with buffer size 100 KB, fastest of 100 * 100
AVX2    encode  9010.17 MB/sec
AVX2	decode	7036.27 MB/sec
plain   encode  1500.42 MB/sec
plain   decode  1561.01 MB/sec
SSSE3	encode	5698.98 MB/sec
SSSE3	decode	3901.21 MB/sec
SSE42	encode	5698.52 MB/sec
SSE42	decode	5103.18 MB/sec
Testing with buffer size 10 KB, fastest of 1000 * 100
AVX2    encode  8928.00 MB/sec
AVX2	decode	6923.62 MB/sec
plain   encode  1504.46 MB/sec
plain   decode  1558.74 MB/sec
SSSE3   encode  5672.21 MB/sec
SSSE3	decode	3912.51 MB/sec
SSE42	encode	5674.36 MB/sec
SSE42	decode	5078.06 MB/sec
Testing with buffer size 1 KB, fastest of 1000 * 1000
AVX2    encode  6726.56 MB/sec
AVX2	decode	5761.74 MB/sec
plain	encode	1420.93 MB/sec
[...]

PR #22, opened 2016-10-16
Rearrange source tree, add SSE4.1, SSE4.2 and AVX code paths

mayeut:
:warning: Updated with latest commits !

The source tree now has an `arch` folder where all architecture specific files reside.

SSE4.1 has been added for encoding.
SSE4.2 has been added for decoding.
AVX has been added for decoding/encoding.

Speed-up on `Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz` using `Apple LLVM
version 8.0.0 (clang-800.0.38)`

SSE4.1 encoding: +5% compared to SSSE3
SSE4.1 decoding: unchanged (same implementation as SSSE3)
SSE4.2 decoding: +29% compared to SSSE3/SSE4.1
SSE4.2 encoding: +5% compared to SSE4.1 (using the same code as SSE4.1 though, compiler is better on this compilation unit)
AVX decoding: +4% compared to SSE4.2
AVX encoding: +7% compared to SSE4.2

Full results:
```
Filling buffer with 10.0 MB of random data...
Testing with buffer size 10 MB, fastest of 100 * 1
AVX2	encode	8086.63 MB/sec
AVX2	decode	7037.83 MB/sec
plain	encode	1499.12 MB/sec
plain	decode	1474.84 MB/sec
SSSE3	encode	5602.90 MB/sec
SSSE3	decode	3884.43 MB/sec
SSE41	encode	5845.86 MB/sec
SSE41	decode	3884.45 MB/sec
SSE42	encode	6201.14 MB/sec
SSE42	decode	5088.64 MB/sec
AVX	encode	6503.20 MB/sec
AVX	decode	5274.49 MB/sec
Testing with buffer size 1 MB, fastest of 100 * 10
AVX2	encode	9693.09 MB/sec
AVX2	decode	7063.94 MB/sec
plain	encode	1500.92 MB/sec
plain	decode	1480.58 MB/sec
SSSE3	encode	5680.65 MB/sec
SSSE3	decode	3899.75 MB/sec
SSE41	encode	5951.01 MB/sec
SSE41	decode	3899.35 MB/sec
SSE42	encode	6321.86 MB/sec
SSE42	decode	5108.12 MB/sec
AVX	encode	6663.97 MB/sec
AVX	decode	5296.59 MB/sec
Testing with buffer size 100 KB, fastest of 100 * 100
AVX2	encode	9769.84 MB/sec
AVX2	decode	7012.76 MB/sec
plain	encode	1500.48 MB/sec
plain	decode	1479.17 MB/sec
SSSE3	encode	5697.31 MB/sec
SSSE3	decode	3901.69 MB/sec
SSE41	encode	5947.30 MB/sec
SSE41	decode	3901.49 MB/sec
SSE42	encode	6327.47 MB/sec
SSE42	decode	5103.40 MB/sec
AVX	encode	6682.34 MB/sec
AVX	decode	5291.32 MB/sec
Testing with buffer size 10 KB, fastest of 1000 * 100
AVX2	encode	9722.94 MB/sec
AVX2	decode	6914.21 MB/sec
plain	encode	1503.13 MB/sec
plain	decode	1476.89 MB/sec
SSSE3	encode	5662.68 MB/sec
SSSE3	decode	3912.20 MB/sec
SSE41	encode	5907.25 MB/sec
SSE41	decode	3916.50 MB/sec
SSE42	encode	6285.56 MB/sec
SSE42	decode	5067.58 MB/sec
AVX	encode	6646.36 MB/sec
AVX	decode	5269.32 MB/sec
Testing with buffer size 1 KB, fastest of 1000 * 1000
AVX2	encode	7200.30 MB/sec
AVX2	decode	5681.45 MB/sec
plain	encode	1420.39 MB/sec
plain	decode	1441.73 MB/sec
SSSE3	encode	5003.45 MB/sec
SSSE3	decode	3611.75 MB/sec
SSE41	encode	5274.16 MB/sec
SSE41	decode	3593.46 MB/sec
SSE42	encode	5466.90 MB/sec
SSE42	decode	4478.80 MB/sec
AVX	encode	5738.58 MB/sec
AVX	decode	4712.87 MB/sec
```


mayeut:
Here are a couple things I don't like in this PR (and are probably in line with your comments from #18):
- Managing only 1 of the functions for a code path (c.f. [codec_choose.c](https://github.com/aklomp/base64/pull/22/files#diff-045c1cf8624fa0c9fe3333f7a29d7f61R187))
- Code duplication for `enc_translate` and `_mm_bswap_epi32`

Would it be OK to move those functions to the `enc` or `dec` folders ?
This way, it would be easy to pick the right flavor by an include depending on the code path (and would allow to add the decoding function with no duplication).

mayeut:
@aklomp, this PR will conflict with #18 at some point. I'd rather take this one as a basis for the other one if that's ok with you. I'll rebase the other one after being done with this one.

aklomp:
In regard to point 1, there is some precedent in regard to messing with codecs: the Neon64 codec [inlines the uint64 codecs](https://github.com/aklomp/base64/blob/master/lib/codec_neon64.c#L57) as an optimization.

Perhaps the code would become less "weird" if you implemented `BASE64_DEC_FUNCTION(sse41)` inside `codec_sse41.c` as a copy of `BASE64_DEC_FUNCTION(ssse3)`. Then you can keep the weird logic out of `codec_choose.c`. I think that the (minor) code duplication is acceptable in that case.

In regard to point 2, duplicating entire functions is something I'd like to avoid. After some thought, I think your proposal of splitting the functions into separate files makes sense. That's how this library handles the architecture-specific stuff in general: `#include`'ing .c files where necessary. Maybe we could make an `enc_translate.ssse3.c` file and a `_mm_bswap_epi32.ssse3.c` file (extensions indicate the SSE compliance level) and include them in both SSSE3 and SSE4.1/4.2 entry points?

Or maybe an `sse_common.h` that contains those functions? Though I think that makes the structure less clear.

mayeut:
> In regard to point 2, duplicating entire functions is something I'd like to avoid. After some thought, I think your proposal of splitting the functions into separate files makes sense. That's how this library handles the architecture-specific stuff in general: #include'ing .c files where necessary. Maybe we could make an enc_translate.ssse3.c file and a _mm_bswap_epi32.ssse3.c file (extensions indicate the SSE compliance level) and include them in both SSSE3 and SSE4.1/4.2 entry points?

My thoughts exactly. I will push a proposal for this.

mayeut:
I pushed a proposal.

aklomp:
I like this proposal, it looks clean. It does add a new layer of indirection and obfuscation until you get to the actual code that is doing the work, perhaps making it a little less approachable for casual observer, but that's the price we pay for not duplicating code. (This library is already quite... bespoke.)

The new setup of the `codec_*.c` files really makes clear that support for an architecture is implemented by chaining together different modules, I like that conceptually.

One comment I have is about the naming of the files/folders. I know that I suggested to put the SSE architecture level in the filename, but I pause at the `common` folder. I wonder if a scheme like this might be cleaner:

```
├── lib
    ├── codec_avx2.c
    ├── codec_choose.c
    ├── codec_neon32.c
    ├── codec_neon64.c
    ├── codec_plain.c
    ├── codec_sse41.c
    ├── codec_ssse3.c
    ├── codecs.h
    ├── arch
    │   ├── sse2
    │   │   └── compare_macros.h
    │   ├── ssse3
    │   │   ├── dec_reshuffle.c
    │   │   ├── enc_reshuffle.c
    │   │   ├── enc_translate.c
    │   │   └── _mm_bswap_epi32.c
    │   └── sse41
    │        └── enc_reshuffle.c
    ├── dec
    │   ├── avx2.c
    │   ├── head.c
    │   ├── neon.c
    │   ├── ssse3.c
    │   ├── tail.c
    │   ├── uint32.c
    │   └── uint64.c
    ├── enc
    │   ├── avx2.c
    │   ├── head.c
    │   ├── neon32.c
    │   ├── neon64.c
    │   ├── ssse3.c
    │   ├── tail.c
    │   ├── uint32.c
    │   └── uint64.c
    ├── exports.txt
    ├── lib.c
    └── lib_openmp.c
```

Maybe not -- `arch` implies that that's where the architecture-specific codecs live, which they don't.

mayeut:
The `common` folder is for code that's not specific to encoding or decoding.

The thing with your proposal, and as you mention at the end, is that architecture specific code also appear in directories which is not `arch`.

That might be using a sledgehammer to crack a nut but what about this then:
```
├── lib
    ├── codec_avx2.c
    ├── codec_choose.c
    ├── codec_neon32.c
    ├── codec_neon64.c
    ├── codec_plain.c
    ├── codec_sse41.c
    ├── codec_ssse3.c
    ├── codecs.h
    ├── common
    │   └── arch
    │       └── x86
    │           ├── sse2
    │           │   └── compare_macros.h
    │           └── ssse3
    │               └── _mm_bswap_epi32.c
    ├── dec
    │   └── arch
    │       ├── generic
    │       │   ├── head.c
    │       │   ├── tail.c
    │       │   ├── uint32.c
    │       │   └── uint64.c
    │       ├── arm
    │       │   └── neon.c
    │       └── x86
    │           ├── ssse3
    │           │   ├── dec_reshuffle.c
    │           │   └── ssse3.c
    │           └── avx2
    │               └── avx2.c
    ├── enc
    │   └── arch
    │       ├── generic
    │       │   ├── head.c
    │       │   ├── tail.c
    │       │   ├── uint32.c
    │       │   └── uint64.c
    │       ├── arm
    │       │   └── neon32.c
    │       ├── arm64
    │       │   └── neon64.c
    │       └── x86
    │           ├── ssse3
    │           │   ├── enc_reshuffle.c
    │           │   ├── enc_translate.c
    │           │   └── ssse3.c
    │           └── avx2
    │               └── avx2.c
    ├── exports.txt
    ├── lib.c
    └── lib_openmp.c
```

mayeut:
or, still heavy, this one:
```
├── lib
    ├── codec_avx2.c
    ├── codec_choose.c
    ├── codec_neon32.c
    ├── codec_neon64.c
    ├── codec_plain.c
    ├── codec_sse41.c
    ├── codec_ssse3.c
    ├── codecs.h
    ├── arch
    │   ├── generic
    │   │   ├── dec
    │   │   │   ├── 32
    │   │   │   │   └── loop.c
    │   │   │   ├── 64
    │   │   │   │   └── loop.c
    │   │   │   ├── head.c
    │   │   │   └── tail.c
    │   │   └── enc
    │   │       ├── 32
    │   │       │   └── loop.c
    │   │       ├── 64
    │   │       │   └── loop.c
    │   │       ├── head.c
    │   │       └── tail.c  
    │   ├── arm
    │   │   ├── dec
    │   │   │   └── neon
    │   │   │       └── loop.c
    │   │   └── enc
    │   │       └── neon
    │   │           └── loop.c
    │   ├── arm64
    │   │   └── enc
    │   │       └── neon
    │   │           └── loop.c
    │   └── x86
    │       ├── common
    │       │   ├── sse2
    │       │   │   └── compare_macros.h
    │       │   └── ssse3
    │       │       └── _mm_bswap_epi32.c
    │       ├── dec
    │       │   ├── ssse3
    │       │   │   ├── dec_reshuffle.c
    │       │   │   └── loop.c
    │       │   └── avx2
    │       │       └── loop.c
    │       └── enc
    │           ├── ssse3
    │           │   ├── enc_reshuffle.c
    │           │   ├── enc_translate.c
    │           │   └── loop.c
    │           └── avx2
    │               └── loop.c
    ├── exports.txt
    ├── lib.c
    └── lib_openmp.c
```

aklomp:
Your proposals are not bad, but like you note, they are quite heavyweight and impose a lot of structure. In the last example, I don't see the need to split by _platform_ (arm, x86..) when splitting by SIMD flavour is sufficient.  In other words, I think neon/neon64/ssse3/avx2 and so on could/should be peers.

Here's a bit more radical design that shakes up the tree a little bit, but keeps the structure flat and relatively compact. I've now moved _all_ arch-sensitive code into the `/arch` directory, including the former `codec_<arch>.c` files.

```
├── lib
    ├── arch
    │   ├── avx2
    │   │   ├── codec.c
    │   │   ├── dec.c
    │   │   └── enc.c
    │   ├── generic
    │   │   ├── codec.c
    │   │   ├── dec
    │   │   │   ├── head.c
    │   │   │   ├── tail.c
    │   │   │   ├── uint32.c
    │   │   │   └── uint64.c
    │   │   └── enc
    │   │       ├── head.c
    │   │       ├── tail.c
    │   │       ├── uint32.c
    │   │       └── uint64.c
    │   ├── neon32
    │   │   ├── codec.c
    │   │   ├── dec.c
    │   │   └── enc.c
    │   ├── neon64
    │   │   ├── codec.c
    │   │   └── enc.c
    │   ├── sse2
    │   │   └── compare_macros.h
    │   ├── sse41
    │   │   ├── codec.c
    │   │   ├── enc.c
    │   │   └── enc_reshuffle.c
    │   ├── sse42
    │   │   ├── codec.c
    │   │   ├── enc.c
    │   │   └── enc_reshuffle.c
    │   └── ssse3
    │       ├── _mm_bswap_epi32.c
    │       ├── codec.c
    │       ├── dec.c
    │       ├── dec_reshuffle.c
    │       ├── enc.c
    │       ├── enc_reshuffle.c
    │       └── enc_translate.c
    ├── codec_choose.c
    ├── codecs.h
    ├── exports.txt
    ├── lib.c
    └── lib_openmp.c
```

What I don't like:
- `enc.c` and `dec.c` are not as emphasized as I would like them to be.
- The name `codec.c` clashes basly with `enc.c` and `dec.c`. Maybe it should be something like `arch.c`, `dispatch.c`, or even `main.c`.
- The `generic` folder has separate `enc` and `dec` folders which the other archs lack. Oh well, fair game.

mayeut:
I kind of like that one.
Regarding your comments, I would probably go with `enc_loop.c` and `dec_loop.c` rather than just  `enc.c` and `dec.c`. This way, I think it does not clash with `codec.c` as well as adding some emphasis.
As for the `generic`, in order to get something that's more similar to other arch, I'd probably go with:
```
├── lib
    ├── arch
    │   ├── avx2
    │   │   ├── codec.c
    │   │   ├── dec_loop.c
    │   │   └── enc_loop.c
    │   ├── generic
    │   │   ├── 32
    │   │   │   ├── dec_loop.c
    │   │   │   └── enc_loop.c
    │   │   ├── 64
    │   │   │   ├── dec_loop.c
    │   │   │   └── enc_loop.c
    │   │   ├── codec.c
    │   │   ├── dec_head.c
    │   │   ├── dec_tail.c
    │   │   ├── enc_head.c
    │   │   └── enc_tail.c
    │   ├── neon32
    │   │   ├── codec.c
    │   │   ├── dec_loop.c
    │   │   └── enc_loop.c
```

aklomp:
Yes, I like this. Renaming to `enc_loop.c` and `dec_loop.c` gives extra emphasis and conveys something about the contents of the file. (Nitpick: maybe `enc_body.c` as the counterpart to `enc_head.c` and `enc_tail.c`? On the other hand, it is the mainloop...)

`generic` is a special case since it's not an architecture and is reused everywhere, so I'm not bothered by the broken symmetry.

`codec.c` is maybe not so bad, all considered. We can always change it later.

And the most important question: will this make it possible to cleanly implement AVX, SSE41 and SSE42? I think it will.

mayeut:
Yes, it will allow to cleanly implement SSE41, SSE42 and AVX !
I will make changes tomorrow based on that latest proposal. I like the `_loop` suffix more than the `_body` one, I think it is more representative of the content at first glance. Please, tell me which one you wan't to go with (and as you said for `codec.c`, it can always be changed later).

aklomp:
I think that `_loop` is probably better, it conveys the intent more accurately. (I.e. it's the SIMD codec's mainloop.) The symmetry with `head` and `tail` is nice but secondary.

I'd make the changes myself if I had more time, but your tempo is hard to beat :)  This is kind of a big one though, so I'd like to take a bit more time on the review.

aklomp:
Hm, looks good to me. Like I noted yesterday, I'm going to take some time with this. So far the only real omission is that the README has not been updated for the new codecs - but that can be done in a final commit.

mayeut:
I added modifications to the `README.md` file in the SSE4.1 commit.
If that looks OK to you, I will do the same for SSE4.2 & AVX commits.
Otherwise let's just iterate on the SSE4.1 modifications.

lib/codec_choose.c
```diff
@@ -148,21 +157,21 @@ codec_choose_arm (struct codec *codec)
 static bool
 codec_choose_x86 (struct codec *codec)
 {
-#if (__x86_64__ || __i386__ || _M_X86 || _M_X64) && (HAVE_AVX2 || HAVE_SSSE3 || HAVE_SSE41 || HAVE_SSE42)
+#if (__x86_64__ || __i386__ || _M_X86 || _M_X64) && (HAVE_AVX2 || HAVE_SSSE3 || HAVE_SSE41 || HAVE_SSE42 || HAVE_AVX)
```
aklomp: If you change the preprocessor stuff to the format above (or something else), you might as well do it in the SSE4.1 commit straight away.

lib/arch/sse42/dec_loop.c
```diff
@@ -0,0 +1,85 @@
...
+	//  #  From       To        Add    Index  Characters
+	//  1  [43]       [62]      +19        0  +
+	//  2  [47]       [63]      +16        1  /
+	//  3  [48..57]   [52..61]   +4  [2..11]  0..9
+	//  4  [65..90]   [0..25]   -65       15  A..Z
+	//  5  [97..122]  [26..51]  -71       14  a..z
+	// (6) Everything else => invalid input
+
+	// LUT:
+	const __m128i lut = _mm_setr_epi8(19,16,4,4,4,4,4,4,4,4,4,4,0,0,-71,-65);
```
aklomp: I'm not a big fan of leaving out the whitespace in the list. If line length becomes an issue I'd prefer to split the list in blocks of four. Example, [lib/arch/ssse3/enc_translate.c](https://github.com/mayeut/base64/blob/sse41/lib/arch/ssse3/enc_translate.c#L5).

lib/arch/sse42/dec_loop.c
```diff
@@ -0,0 +1,85 @@
...
+	if (_mm_cmpistrc(range, str, _SIDD_UBYTE_OPS | _SIDD_CMP_RANGES | _SIDD_NEGATIVE_POLARITY)) {
+		break;
+	}
+
+	// Compute indices for table look up:
+	// First indices for ranges #[1..3]. Others are invalid.
+	__m128i indices = _mm_subs_epu8(str, _mm_set1_epi8(46));
+
+	// Compute mask for ranges #4 and #5:
+	__m128i maskUL = CMPGT(str, 64);
```
aklomp: Maybe now is the time to ask; what is the meaning behind `maskUL` and `maskL`? Unsigned Long and Long? The library generally doesn't use uppercase in variable names, unless there is a compelling reason.

lib/codec_choose.c
```diff
@@ -49,6 +49,9 @@
 #ifndef bit_SSE42
 #define bit_SSE42 (1 << 20)
 #endif
+#ifndef bit_AVX
+#define bit_AVX (1 << 28)
+#endif
 
 #define bit_XSAVE_XRSTORE (1 << 27)
```
aklomp: I know this is not your code, but I was looking at the [Wikipedia page for CPUID](https://en.wikipedia.org/wiki/CPUID) and it looks like it might be better to use bit 26 instead of 27 to check for AVX. Bit 26 is whether the hardware supports `XSAVE` and `XRESTOR`, bit 27 is whether the OS supports `XRESTOR`. If you agree, this commit would be the place to change it. Compiling with bit 26 works fine for me. (Question: how can an OS influence the result of `CPUID`?)

lib/codec_choose.c
```diff
@@ -148,21 +157,21 @@ codec_choose_arm (struct codec *codec)
...
 codec_choose_x86 (struct codec *codec)
 {
-#if (__x86_64__ || __i386__ || _M_X86 || _M_X64) && (HAVE_AVX2 || HAVE_SSSE3 || HAVE_SSE41 || HAVE_SSE42)
+#if (__x86_64__ || __i386__ || _M_X86 || _M_X64) && (HAVE_AVX2 || HAVE_SSSE3 || HAVE_SSE41 || HAVE_SSE42 || HAVE_AVX)
 
 	unsigned int eax, ebx = 0, ecx = 0, edx;
 	unsigned int max_level;
 
-	#ifdef _MSC_VER 
+	#ifdef _MSC_VER
```
aklomp: niiiice :)

lib/arch/sse42/dec_loop.c
mayeut: I will do split in blocks of four.

aklomp:
Thanks, this is really high quality work. I had some small comments, but for the rest this is ready to merge. This commit will turn this library into super abstract postmodern cubist art :)

In regards to the README, I think the way that the `CFLAGS` for the different x86 SIMD families are presented could be made more clear. Now it's in a long sentence, but perhaps a table is more clear.

| X86 SIMD family | Compiler CFLAGS |
|-------|-------|
| SSSE 3 | `SSSE3_CFLAGS=-mssse3` |
| SSE 4.1 | `SSE41_CFLAGS=-msse4.1` |
| SSE 4.2 | `SSE42_CFLAGS=-msse4.2` |
| AVX | `AVX_CFLAGS=-mavx` |
| AVX2 | `AVX2_CFLAGS=-mavx2` |

NEON wouldn't fit in this table, that can remain a special case.

lib/codec_choose.c
mayeut: The wikipedia states it is bit 28 for `Advanced Vector eXtensions` (modified case on `eXtensions` on purpose).
Bit 26 & 27 are `XSAVE/XRESTOR` & `XSAVE enabled by OS`.
The mechanism to check for `AVX` is the same as `AVX2` except for the `AVX` flag instead of `AVX2`. Meaning all the stuff around checking OS support, calling `xgetbv`is the same.
Yes, the OS can have an influence on `CPUID`. If I remember correctly (this could be checked in the [Intel® 64 and IA-32 Architectures Software Developer Manuals](http://www.intel.com/content/www/us/en/processors/architectures-software-developer-manuals.html)), when the OS/kernel sets-up the CPU, it configures specific registers telling it that it will use `XSAVE/XRESTORE` and supports saving `ymm` registers. The first reflects in `CPUID`and the latter in the call to `xgetbv`.

aklomp: I completely missed bit 28, I must be going blind. Wikipedia informs us that bit 5 in `EAX=7` should be the `avx2` bit. I believe the current code was based on [this](https://software.intel.com/en-us/articles/how-to-detect-new-instruction-support-in-the-4th-generation-intel-core-processor-family) Intel snippet or a derivative.

mayeut:
I made modifications you asked for in the SSE4.2 commit. 
In order to ease the process for `README.md`, do you think I can go on like the one on `SSE4.1` for the 2 remaining commits and, then, let you commit the table you propose ?

lib/codec_choose.c
mayeut: Sorry, I missed that one. Since it can impact the same file for every commit, I'd rather have this done after if possible.

aklomp:
Regarding the README: sure, let's make it _complete_ now and worry about improving it later. The scope of this pull request is already quite large.

Regarding the `BASE64_X86_SIMD` simplification: that one can also be done later. I'm thinking of doing a small cleanup round after the dust has settled to fix some syntactic cruft, and this would be on the list.

mayeut:
All done !

aklomp:
Banzai! Merged. Thanks for a great contribution.

PR #24, opened 2016-10-20
Add arm64, armv7hf and ppc64 builds to travis-ci

mayeut:
Uses docker to build and test neon32/neon64 code paths as well as big-endian plain code path.
It uses docker images I prepared for [libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo).
Docker files I used to push those to docker hub are [here](https://github.com/mayeut/ljtbuilder)

Just thought I'd share the idea.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,38 @@
 language: c
 
-compiler:
-  - clang
-  - gcc
+matrix:
+  include:
+    - os: linux
+      compiler: clang
+      env: SSSE3_CFLAGS=-mssse3 SSE41_CFLAGS=-msse4.1 SSE42_CFLAGS=-msse4.2 AVX_CFLAGS=-mavx
+    - os: linux
+      compiler: gcc
+      env: SSSE3_CFLAGS=-mssse3 SSE41_CFLAGS=-msse4.1 SSE42_CFLAGS=-msse4.2 AVX_CFLAGS=-mavx
+    - os: linux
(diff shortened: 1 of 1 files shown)


mayeut:
PS: have a look at the travis-ci check which does include those builds.

aklomp:
Nice, we definitely need CI for the ARM targets. My obvious response to this pull request would be that we introduce a dependency on your images, but then again I think the trade-off is worth it. (Better than *not* having ARM CI.)

In your second comment, are you alluding to the fact that the NEON performance is worse than the generic codec? Makes sense if we're emulating through QEMU.

mayeut:
No, the second comment was to make sure you saw those builds. Performance is "of course bad" with QEMU.

The comment about using my images was in fact to make sure the dependency "issue" was seen.
Maybe you'd wan't to get those on your end to make sure they never disappear. I won't delete them though. I know that `libjpeg-turbo` maintainer will manage those images in his own repo when he gets around it.

mayeut:
Added ppc64 in order to test on big-endian architecture.

mayeut:
This has no use anymore.

aklomp:
Damn, I completely forgot about this PR.  I spent much of last week learning about the CI and polishing my CI branch, to find out now that you already contributed much the same thing... three years ago. I definitely feel like an idiot now. Well, at least the past week was a good learning experience for me.

Thanks for your contributions to this library, and sorry for glancing past this PR.

PR #28, opened 2016-10-22
Fix NEON64 build with gcc

mayeut:
gcc does not understand the `vld1q_u8_x4` intrinsic.
Replace with 4 x `vld1q_u8`.

Found that out while rebasing #24

### DIFF
--- a/lib/arch/neon64/codec.c
+++ b/lib/arch/neon64/codec.c
@@ -16,29 +16,25 @@
 
 #define CMPGT(s,n)	vcgtq_u8((s), vdupq_n_u8(n))
 
-// Encoding
-// With this transposed encoding table, we can use
-// a 64-byte lookup to do the encoding.
-// Read the table top to bottom, left to right.
-static const char *base64_table_enc_transposed =
+static inline uint8x16x4_t
+load_64byte_table (const uint8_t *p)
 {
-	"AQgw"
-	"BRhx"
-	"CSiy"
-	"DTjz"
-	"EUk0"
-	"FVl1"
-	"GWm2"
-	"HXn3"
-	"IYo4"
-	"JZp5"
-	"Kaq6"
-	"Lbr7"
-	"Mcs8"
-	"Ndt9"
-	"Oeu+"
-	"Pfv/"
-};
+#if defined(__GNUC__) && !defined(__clang__)
+	// As of October 2016, GCC does not support the 'vld1q_u8_x4()' intrinsic.
+	uint8x16x4_t ret;
+	ret.val[0] = vld1q_u8(p +  0);
+	ret.val[1] = vld1q_u8(p + 16);
+	ret.val[2] = vld1q_u8(p + 32);
+	ret.val[3] = vld1q_u8(p + 48);
+	return ret;
+#else
+	return vld1q_u8_x4(p);
+#endif
+}
+
+// Encoding
+// Use a 64-byte lookup to do the encoding.
+// Reuse existing base64_table_enc table.
 
 // Decoding
 // The input consists of five valid character sets in the Base64 alphabet,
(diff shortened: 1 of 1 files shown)


aklomp:
Hm, I think it would be better to insert a GCC shim, unless your solution generates the same machine instructions as the intrinsic. (I don't have immediate access to an AArch64 compiler right now, so I can't check that.) I would imagine that compilers would recognize the pattern and optimize, but better to check.

The GCC shim could look something like this (pseudocode):

```c
#if defined(__GCC__) && GCC_VERSION < 60100
static inline uint8x16x4_t
vld1q_u8_x4 (const uint8_t *p)
{
        uint8x16x4_t ret;

        ret.val[0] = vld1q_u8(p +  0);
        ret.val[1] = vld1q_u8(p + 16);
        ret.val[2] = vld1q_u8(p + 32);
        ret.val[3] = vld1q_u8(p + 48);

        return ret;
}
#endif
```

Or with inline assembly in case we need to generate a custom instruction (I try to avoid this in this lib, but sometimes you have to):

```c
#if GCC_VERSION < 60100
static inline uint8x16x4_t
vld1q_u8_x4 (const uint8_t *a)
{
        uint8x16x4_t ret;

        __asm __volatile("vld1q.8 ... <magic>",
                : ...
                : ret
                : ... );

        return ret;
}
#endif
```

mayeut:
It does not generate the same instructions (I would have guessed that it should have been optimized the same way but this is not the case...). But this is not in the critical performance path, it's outside the loop. Using assembly here may be a little too much. Even though the instruction is very simple for this intrinsic `ld1.16b { v0, v1, v2, v3 }, [x8]`, getting the proper register sequence can be a pain (furthermore using a shim, which will always reuse the same register sequence, thus, 4 copies are needed afterwards). I can add the first shim (without assembly) for GCC if you want, but I'm not sure it's worth it.

aklomp:
Hm, you're right that this is not in the critical path. However, by creating the GCC shim, we get the added benefit of being able to use it for the table lookup in the encoding function. We can then replace the arch-specific, column-major encoding LUT with the generic one from `lib/lib.c`.

The other option is to use the same method for loading tables as the encoding function: reorder the LUT to column-major order and use `vld4q_u8()`.

Else, a cleaner middle way is to not put the GCC shim in a separate function but to inline it:

```c
uint8x16x4_t tbl_dec1;
uint8x16x4_t tbl_dec2;

// As of October 2016, GCC does not support the `vld1q_u8_x4()' intrinsic.
#ifdef __GCC__
        tbl_dec1.val[0] = vld1q_u8(base64_dec_lut1 +  0);
        tbl_dec1.val[1] = vld1q_u8(base64_dec_lut1 + 16);
        tbl_dec1.val[2] = vld1q_u8(base64_dec_lut1 + 32);
        tbl_dec1.val[3] = vld1q_u8(base64_dec_lut1 + 48);

        tbl_dec2.val[0] = vld1q_u8(base64_dec_lut2 +  0);
        tbl_dec2.val[1] = vld1q_u8(base64_dec_lut2 + 16);
        tbl_dec2.val[2] = vld1q_u8(base64_dec_lut2 + 32);
        tbl_dec2.val[3] = vld1q_u8(base64_dec_lut2 + 48);
#else
        tbl_dec1 = vld1q_u8_x4(base64_dec_lut1);
        tbl_dec2 = vld1q_u8_x4(base64_dec_lut2);
#endif
```

aklomp:
If I have to take a position, I'd choose the GCC shim without inline assembly, and also use that function in the encoder.

Maybe it would be altogether cleaner to introduce a `load_64byte_table()` function (name is just a suggestion) that encapsulates the compiler-specific stuff:

```c
static inline uint8x16x4_t
load_64byte_table (const uint8_t *tbl)
{
#ifdef __GCC__
        uint8x16x4_t ret;

        ret.val[0] = vld1q_u8(tbl +  0);
        ret.val[1] = vld1q_u8(tbl + 16);
        ret.val[2] = vld1q_u8(tbl + 32);
        ret.val[3] = vld1q_u8(tbl + 48);

        return ret;
#else
        return vld1q_u8_x4(tbl);
#endif
}
```

mayeut:
OK, I will go that way then.

mayeut:
Done.

aklomp:
Merged, thanks! I couldn't find evidence of a CI test for this build, but I trust that you tested the code.

mayeut:
The last commit wasn't tested with CI (I lost track of what I tested with CI) but was tested.
Anyway #24 travis-ci check is now back to `successful check`

Issue #29, opened 2016-10-23
Issues with recent changes


mscdex:
After pulling in recent changes I encountered a couple of issues so far:

1. With the new AVX target, I found that the XRSTORE check only works for AVX2. Here's what I ended up using in codec_choose.c to support AVX (not sure how correct this is):

```c
if (max_level >= 7) {
  #ifdef HAVE_AVX2
    __cpuid_count(7, 0, eax, ebx, ecx, edx);
  #endif
  #ifdef HAVE_AVX
  if (!(ecx & bit_XSAVE_XRSTORE)) {
    __cpuid(1, eax, ebx, ecx, edx);
  }
  #endif

  // ...
```

2. My build setup uses libtool, which apparently has issues on macOS when multiple files have the same base name, in this case it's the codec.c files in each target/arch directory. Supposedly Windows may have a similar issue in some cases?

mayeut:
Well the code for AVX2 was buggy before. It's still buggy now... I will make a PR to get this fixed.

mayeut:
Regarding you second bullet point, without any pointer to your configuration, it will be hard to provide any support. I'm on macOS, I have lib tool installed, I can try to check that if you provide me with your fork URL.

aklomp:
So the AVX detection was fixed (thanks @mayeut), which leaves the macOS libtool problem. @mscdex, could you provide some more information about what exactly is going wrong, such as terminal output or error messages?

PR #30, opened 2016-10-23
Fix AVX/AVX2 runtime detection

mayeut:
XSAVE_XRSTORE/AVX2/AVX bits were checked after a `CPUID.(EAX=07H,
ECX=0H)` call.
Only AVX2 flag shall be checked after that call.
Check for XSAVE_XRSTORE/AVX shall be done after a `CPUID.(EAX=01H,
ECX=0H)` call.

Update #29


mscdex:
I can confirm this at least works for AVX now. Thanks! :dancer:

Issue #27, opened 2016-10-21
Invalid input data handling


mayeut:
I'm not sure how "invalid" input data shall be processed. Could this be clarified ?

A few samples using `base64_decode`, what's the expected output for those ?
- "Zm9vYg=" decodes successfully though it's not properly terminated
- "Zm9vYg" decodes successfully though it's not properly terminated
- "Zm9vY" decodes successfully though it misses data
- "Zm9vYmF=Zm9v" decodes successfully though it contains too much data or corruption occurred.

This could be extended to streaming API, but I thought I'd start with the simple block decode.

aklomp:
The compliance baseline for this library should be the relevant [RFC 4648](https://tools.ietf.org/html/rfc4648). I never liked [Postel's law](https://en.wikipedia.org/wiki/Postel%27s_law) much ("be strict in what you send, and liberal in what you accept"). Being liberal just invites abuse. All your examples should throw an error when they violate the preconditions.

mayeut:
I'll check the `RFC 4648`and try to make a PR to update the test suite and library to handle the invalid properly.

aklomp:
Thanks! Let's have a library that can at least strictly validate. If at any later moment we want to relax the validation, we could create a flag for it.

(Personally, I think that any string that violates the base64 spec as given in the RFC, is not a valid base64 string, and should not be given the benefit of the doubt.)

aklomp:
Maybe one comment: it's quite common for base64 input to contain whitespace (newlines and such), and it would be helpful to have a flag to tell the library to ignore it. This does not negate the fact that a bona-fide base64 string with such characters is invalid, but it would indicate that it would be useful to have a flag that tells the library to strip out those characters before processing. (See also #15).

mayeut:
So, not much else than what you're saying after reviewing `RFC 4648`.

First step, report all violations.
Second step, allow whitespace.

mayeut:
For now, I added validation for `base64_decode` function only.
The streaming implementation would require a final step (as can be seen in validation for `base64_decode`). This requires adding a new function in the API or asking integrators to do the same check as what's done for `base64_decode`.
I'm more in favor of adding a new API function. Could be something like:
```
int
base64_stream_decode_final (struct base64_state *state)
{
	state->eof = BASE64_EOF;
	if (state->bytes == 0) {
		return 1;
	}
	return 0;
}
```

aklomp:
[...]

PR #32, opened 2016-11-06
Fix invalid input data handling for base64_decode

mayeut:
Fix invalid input data handling for base64_decode function and add tests for those.

Update #27

### DIFF
--- a/lib/arch/generic/dec_tail.c
+++ b/lib/arch/generic/dec_tail.c
@@ -29,13 +29,16 @@
 			break;
 		}
 		if ((q = base64_table_dec[*c++]) >= 254) {
+			st.bytes++;
 			// When q == 254, the input char is '='.
 			// Check if next byte is also '=':
 			if (q == 254) {
 				if (srclen-- != 0) {
+					st.bytes = 0;
 					// EOF:
 					st.eof = BASE64_EOF;
-					ret = (base64_table_dec[*c++] == 254) ? 1 : 0;
+					q = base64_table_dec[*c++];
+					ret = ((q == 254) && (srclen == 0)) ? 1 : 0;
 					break;
 				}
 				else {
@@ -46,7 +49,6 @@
 				}
 			}
 			// If we get here, there was an error:
-			ret = 0;
 			break;
 		}
 		*o++ = st.carry | (q >> 2);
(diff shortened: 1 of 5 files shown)


lib/arch/generic/dec_tail.c
```diff
@@ -5,6 +5,7 @@
 		if ((q = base64_table_dec[*c++]) >= 254) {
 			st.eof = BASE64_EOF;
 			// Treat character '=' as invalid for byte 0:
+			ret = 0;
```
aklomp: Is this strictly necessary when `ret` defaults to 0? (I.e. we assume failure and must "prove" success.)

lib/lib.c
```diff
@@ -184,5 +185,10 @@ base64_decode
 	base64_stream_decode_init(&state, flags);
 
 	// Feed the whole string to the stream reader:
-	return base64_stream_decode(&state, src, srclen, out, outlen);
+	ret = base64_stream_decode(&state, src, srclen, out, outlen);
+	// If when decoding a whole block, we're still waiting for input then fail:
```
aklomp: Could you please insert a blank line before this comment? :)

aklomp:
Excellent work, thanks. More rigorous error checking *and* extra testcases, nice.

lib/arch/generic/dec_tail.c
mayeut: it's not strictly necessary. I find it easier to understand that way, but, since it's just a personal preference, I removed it.

lib/lib.c
mayeut: Done

aklomp:
Did you also take into account the `base64_decode_openmp` function in `lib/lib_openmp.c`? It's a relative of `base64_decode` that splits the work over multiple threads. At the end it merges the results from the threads and then runs a single-threaded `base64_stream_decode` to handle leftover bytes. It should also check whether there are no spurious bytes after the EOF marker, because the testcases fail when `OMP_THRESHOLD == 0` (i.e. always use OpenMP).

mayeut:
Well, I didn't test nor take into account OpenMP. It's not available under Xcode.
I will try to build this using gcc to see what's going on.

aklomp:
Changing the last lines of `base64_decode_openmp` to:

```c
if (result && (state.bytes == 0)) {
        return result;
}

return 0;
```

...fixes the issue (the tests pass).

mayeut:
yes. It fixes that one (pushed it).
There's still an issue if EOF is reached in the OpenMP loop. This could be checked by comparing the number of output bytes vs. expected number of output bytes.

PR #35, opened 2017-02-12
Implement various optimizations

mayeut:
Use Nick Galbreath (@client9) implementation for scalar decoding.
Use Wojciech Mula (@WojciechMula) implementation for AVX2 / SSSE3 with
decoding trick by @aqrit

SSE4.1 & SSE4.2 are now useless.

Speed-up on i7-4870HQ @ 2.5 GHz (clang-800.0.42.1, x86_64)
Plain decoding: +97%
SSSE3 encoding: +13%
SSSE3 decoding: +79%
AVX encoding: +6%
AVX decoding: +57%
AVX2 encoding: +3%
AVX2 decoding: +69%

PR #36, opened 2017-02-14
Plain decoding optimization

mayeut:
This now uses a modified implementation of Nick Galbreath (@client9)
algorithm.
Modifications of the original algorithm include:
- Unaligned access is now optional (macro to enable this)
- Invalid characters are mapped to 0xffffffff instead of 0x01ffffff.
This removes the need to shift by 8 the returned value of the tables on
big-endian architecture with fast unaligned access at the expense of a
different check for invalid character depending on endianness

Speed-up using clang-800.0.42.1 on i7-4870HQ @ 2.5 GHz or iPhone SE:
- x86_64: +97%
- i386: +60%
- arm64: +0%
- armv7: +24%

As a side note, it seems that the iPhone SE processor (Apple A9) has
fast unaligned access which gives even higher speed-up with
corresponding macro set to 1 (+23% / +56%). This may not be the case
for some other ARM processors.

This is the first PR of a series as requested in #35


lib/arch/generic/codec.c
```diff
@@ -19,10 +19,6 @@ BASE64_ENC_FUNCTION(plain)
 BASE64_DEC_FUNCTION(plain)
 {
 	#include "dec_head.c"
-#if BASE64_WORDSIZE == 32
-	#include "32/dec_loop.c"
-#elif BASE64_WORDSIZE == 64
-	#include "64/dec_loop.c"
-#endif
+	#include "dec_loop.c"
 	#include "dec_tail.c"
```
aklomp: Hmm. There's an underlying assumption here that the wordsize is always going to be >= 32 bit. The current code has no such assumption and will work on an 8-bit micro (untested, but in principle...)

Why not retain the `#if BASE64_WORDSIZE >= 32`?

lib/arch/generic/dec_loop.c
```diff
@@ -0,0 +1,46 @@
+// Read source 4 bytes at a time
+// Since we might be writing one byte more than needed,
+// we need to make sure there will still be some room
+// for one extra byte in o.
+// This will be the case if srclen > 0 when the loop
+// is exited
```
aklomp: Hmm, is this also true when the remaining bytes are `=` or `==`?

Makefile
```diff
@@ -63,15 +69,24 @@ lib/libbase64.o: $(OBJS)
...
+	@echo "#define HAVE_SSE42                 $(HAVE_SSE42)"                 >> $@
+	@echo "#define HAVE_AVX                   $(HAVE_AVX)"                   >> $@
+	@echo "#define HAVE_FAST_UNALIGNED_ACCESS $(HAVE_FAST_UNALIGNED_ACCESS)" >> $@
 
-lib/codec_choose.o: lib/config.h
+lib/table_generator: lib/table_generator.o
+	$(CC) $(CFLAGS) -o $@ $^
+
+lib/tables.h: lib/table_generator
+	./lib/table_generator > $@
```
aklomp: I think I'd like to commit the tables as well, as they are (important) data and should be under version control. The program to generate them should also be available in the source distribution. A middle ground could be to leave this line in the Makefile, but *also* commit lib/tables.h. But we can implement that later. (No action needed on this comment.)

aklomp:
Thanks, this is a nice pull request! I left three review notes. If those are worked out, I think this is fit to merge.

lib/arch/generic/dec_loop.c
mayeut: This is true given the contract in the API. The output buffer size must be at least 3/4 the size of the input buffer.

lib/arch/generic/codec.c
mayeut: Right, will change that

Makefile
mayeut: Tables are committed.
I don't know how to not build the table_generator. tables.h is now generated only if the source for the generator is modified.

mayeut: OK, found a way to do that.

mayeut:
All review notes addressed.

aklomp:
Sorry for the delays in responding. Last week was busy personally and professionally. I'll try to move things forward again.

Makefile
```diff
@@ -63,15 +71,25 @@ lib/libbase64.o: $(OBJS)
...
+	@echo "#define HAVE_AVX                   $(HAVE_AVX)"                   >> $@
+	@echo "#define HAVE_FAST_UNALIGNED_ACCESS $(HAVE_FAST_UNALIGNED_ACCESS)" >> $@
+
+lib/table_generator: lib/table_generator.o
+	$(CC) $(CFLAGS) -o $@ $^
+
+lib/tables.h: lib/table_generator.c
+	@$(MAKE) -f $(THIS_FILE) lib/table_generator
+	./lib/table_generator > $@
+
```
aklomp: Isn't this a bit complex? The following rule also works:

```make
lib/tables.h: ./lib/table_generator
	$^ > $@
```

lib/arch/generic/codec.c
```diff
@@ -19,10 +19,8 @@ BASE64_ENC_FUNCTION(plain)
 BASE64_DEC_FUNCTION(plain)
 {
 	#include "dec_head.c"
-#if BASE64_WORDSIZE == 32
+#if BASE64_WORDSIZE >= 32
```
aklomp: This raises the question of whether it would be worthwhile to add the same table-based implementation for 64-bit words. I don't think so because the tables would get very big. (No action needed on this comment.)

lib/table_generator.c
```diff
@@ -0,0 +1,183 @@
...
+	printf("\n};\n");
+}
+
+int main(int argc, char** argv)
+{
+	uint32_t x;
+	uint32_t i = 0;
+	uint32_t ary[256];
+
+	/*  over-ride standard alphabet */
```
aklomp: Uh, so this is meant to create tables to decode "web" Base64? Could be better documented. OK, leave it for later.

aklomp:
Apart from the comment on the Makefile, this pull request looks OK and I'm willing to pull it. Do you want to fix the Makefile? I can also fix it in a followup commit if it saves you time.

lib/table_generator.c
mayeut: Yes, this was in Nick Galbreath original implementation. I left it as is since I only did required modifications to the file.

lib/arch/generic/codec.c
mayeut: I tested an implementation with 64-bit words, it wasn't worth it.

Makefile
mayeut: Since `lib/tables.h` is now under source control, I wanted to generate a new one only if `lib/table_generator.c` was modified. This can be modified to overwrite `lib/tables.h` "each" time libbase64 is built as you suggest.

aklomp: Having `lib/tables.h` in source control is fine, the file doesn't need to be regenerated every time the library is rebuilt.

I think the machinery to force a dependency on `lib/table_generator.c` is a bit too complex. Since `lib/table_generator` already depends on `lib/table_generator.o`, won't the binary always get rebuilt when `lib/table_generator.c` is changed?

If someone wants to regenerate `lib/tables.h`, they can delete that file and it will be rebuilt from source using whatever fresh version of `lib/table_generator.c` is available. I think that's sufficiently straightforward and clean.

mayeut: I've been rethinking a bit about this.
Writing
```Makefile
lib/tables.h: ./lib/table_generator
	$^ > $@
```
will make `lib/tables.h` to be rebuilt every time (well at least on fresh checkout or source tar.gz) since it depends on `lib/table_generator` which is not under source control and thus must be rebuilt.
This might be acceptable but won't allow for cross-compilation with an unpatched makefile since `lib/table_generator` won't be able to run on the build system. The option currently implemented in the PR would allow for cross-compilation out of the box.
The only time `lib/tables.h`and `lib/table_generator` would be rebuilt is if `lib/table_generator.c` changes and in this case, we can assume someone doing this knows what he is doing (i.e. knows cross-compilation won't work in that specific case).
I admit it's a bit more complex but allows for cross-compilation out of the box.
Your choice.

aklomp: Hmm, yes you're right of course. My proposed change would force a rebuild after a fresh checkout, which is the thing we wanted to avoid by putting the tables under source control...

The reason why I proposed the change was that I think the `$(THIS_FILE)` construction is an ugly hack. Can we avoid it in any way? The following thing works (does not rebuild needlessly), at the cost of repeating the recipe for `%.o`, but at the gain of getting rid of the rule for `lib/table_generator`:

```make
lib/tables.h: ./lib/table_generator.c
        $(CC) $(CFLAGS) -o lib/table_generator $^
        ./lib/table_generator > $@
```

mayeut: You're right about the ugly hack but that's the only way I could do this without repeating the recipe for `%.o`as you mentioned.
What you propose is probably a good middle ground. I'll go with that.

mayeut:
Makefile has been modified according to latest review.

aklomp:
Merged, thanks!

PR #41, opened 2017-08-08
Updated benchmarks on i7-4770 @ 3.4 GHz DDR1600

htot:
I have done some benchmark updating today on i7-4770. Also some new benchmarks on Edison x86-64 (this is running Linux 4.13.0-rc1-next x86_640.


aklomp:
Thanks! Merged into master.

Issue #40, opened 2017-05-24
Window x64 platform base_encode outlen is ZERO


tomcatter:
I compiled the library on Windows x64 platform, I specified the sse3 instruction set, then I used it for encode to base64, but the length of the base64 string is zero. I tried the x86, it is working well.

aklomp:
An output length of zero is a strong indication that the codec is not properly compiled in. Are you using the `BASE64_FORCE_SSSE3` flag to force the codec?

Could you compile and run the test suite and print the output?

tomcatter:
Yes, I use this version of base64-0.3.0 release. In the config.h set the `HAVE_SSSE3` to 1.

aklomp:
Hey, could you please re-test with the latest master branch? The merge of issue #42 may have fixed the bug.

aklomp:
As of `v0.5.2`, this library is now built and tested in CI on Windows, by building it with a variety of MSVC and MSYS2/MinGW setups. A number of bugs were also fixed in regard to platform detection and Windows compatibility. Does the bug still happen with the latest release?

I am tempted to close this issue because there is now comprehensive CI coverage for Windows.

PR #42, opened 2017-08-31
Do the same checks in codecs than in codec_choose

mayeut:
MSVC compiler does not define those `__SIMD__` (e.g. `__AVX__`) predefined macros. 

Conditional code is now only dependent on config.h values in all `codec.c` files except for NEON which also includes the same checks as in `codec_choose.c`.

This fixes mayeut/pybase64#8 and probably aklomp/base64#40

### DIFF
--- a/lib/arch/avx/codec.c
+++ b/lib/arch/avx/codec.c
@@ -5,7 +5,7 @@
 #include "../../../include/libbase64.h"
 #include "../../codecs.h"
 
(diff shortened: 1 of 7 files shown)


aklomp:
Merged, thanks. Personally I liked the idea of trusting the compiler's builtin platform vars over the stuff that the user puts on the commandline, but I can see how it fails to be portable.

PR #43, opened 2017-09-18
SSSE3 decoder: rewrite for a couple less instructions

aqrit:
2 instructions fewer than the current version.

I was unable to run the benchmark on Windows, though.

sanity check:
https://gist.github.com/aqrit/a2ccea48d7cac7e9d4d99f19d4759666

code golf forever! @WojciechMula


WojciechMula:
@aqrit Just skimmed through your commit and can say you did something weird, congratulations. :) Is it significantly faster? @lemire

lemire:
I'd be interested in a benchmark.

htot:
I will have some fresh benchmarks soon on i7 and edison (tesed with code from commit 0a69845). Spoiler alert: on i7 we are already memory bandwidth limited. We might be able to speed by making more efficient use of the cache, i.e. not waste the cache on cache the data buffer, but making all fixed data tables are loaded into the cache. On Edison we are CPU limited (slow 500MHz Atom, with relatively fast memory), so we might expect a little improvement there.

aqrit:
This PR is probably a step backwards on Atom because pshufb is slow there.

aklomp:
Nice work, this does appear to be slightly faster in my benchmarks, and it passes the test suite. Normally I'd merge it by that metric alone. However, maintaining a library takes more than just clever code, so I have to put on my schoolmaster/reviewer hat and note:

- The commit is a bit messy. The commit title is uninformative, there are whitespace errors, and there's a comment which was started but not finished. I pushed a branch called 'aqritm' which has a commit that I "fixed" according to how I think it should look.
- More importantly: _I do not understand anything that's going on_ in this commit. Of course it may be because I'm just not clever enough, but from this comment thread it looks like I'm not the only one. The commit deletes all the exposition about hash tables and such, and replaces it with extremely clever code presented in a style of "deal with it". So we have code that passes the benchmarks and the tests, and that I'm still hesitant to merge because it's fundamentally not maintainable for me... One example are the four constant lookup tables. They might have well been presented by God on stone tablets written in gold, because I have _no way_ of verifying or reproducing their contents.

Is it perhaps possible to add some comments to explain the basic ideas behind this code and why it works the way it does? As the maintainer I have to ask that question, because there is more to programming than just cleverness (although an important part). That aside, again props for the achievement.

htot:
@aqrit a previous commit already regressed the performance on Atom, see my PR #44 any idea what causes that exactly?

aqrit:
delta_hash for 0 .. 7F
```
1 1 1 1 1 1 1 1 1 1 1 1 1 8 1 8
2 2 2 2 2 2 2 2 2 2 2 2 2 9 2 9
3 3 3 3 3 3 3 3 3 3 3 3 3 A 3 A
4 4 4 4 4 4 4 4 4 4 4 4 4 B 4 B
5 5 5 5 5 5 5 5 5 5 5 5 5 C 5 C
6 6 6 6 6 6 6 6 6 6 6 6 6 D 6 D
7 7 7 7 7 7 7 7 7 7 7 7 7 E 7 E
8 8 8 8 8 8 8 8 8 8 8 8 8 F 8 F
```

so the mapping is:
3 = plus (0x2B)
A = slash (0x2F)
B = equal sign (0x3D)
4 = numbers (0x30 .. 0x39)
5 6 C = upper case letters (0x41 .. 05A)
7 8 E = lower case letters (0x61 .. 0x7A)

===========================

check_hash for 0 .. 7F
```
7 1 1 1 1 1 1 1 1 1 2 4 6 6 6 8
8 2 2 2 2 2 2 2 2 2 3 5 7 7 7 9
9 3 3 3 3 3 3 3 3 3 4 6 8 8 8 A
A 4 4 4 4 4 4 4 4 4 5 7 9 9 9 B
B 5 5 5 5 5 5 5 5 5 6 8 A A A C
C 6 6 6 6 6 6 6 6 6 7 9 B B B D
D 7 7 7 7 7 7 7 7 7 8 A C C C E
E 8 8 8 8 8 8 8 8 8 9 B D D D F
```

so the mapping is:
0 1 2 3 B D F = invalid
4 = invalid below 31
5 = invalid below 41
6 = invalid below 2B
7 = invalid below 5A
8 = invalid below 4B
9 = invalid below 7A
A = invalid below 2F
C = invalid below 4F
E = invalid below 6F

for 80 .. FF (signed input):
the check_value will either be signed or 0
so...
_mm_adds_epi8( signed, signed) = signed 
_mm_adds_epi8( 0, signed) = signed

===============

> regressed the performance on Atom... any idea?

3 pshufb + pmovmskb = ~20 cycles ?
I  wonder if the new packing code is slower on old hardware as well.
It trades out logic/shift for mul.

aqrit:
comments added, tried to keep them to just 2 pages worth.

aklomp:
On the three platforms where I've benchmarked this code, performance seems equal to the previous code. Is there a platform where performance is significantly enhanced?

aqrit:
> Nice work, this does appear to be slightly faster in my benchmarks

> On the three platforms where I've benchmarked this code, performance seems equal to the previous code. Is there a platform where performance is significantly enhanced?

significantly? No.
It was slightly faster on my Westmere.
The newer CPUs only have one shuffle port, so it might bottleneck on that.

htot:
I can try this on NUC baytrail / edison (both silvermont). Maybe this weekend.

aklomp:
Yeah, I also noticed my own inconsistency there :smile_cat: That's why I asked - I retested this codec this morning on three machines I had at hand (i5 4590S, Atom N270 and an old Xeon) and could not find noticeable improvement compared to master. Maybe master is benefiting from two years of compiler advancements?

As said before, I'm not warming much to this algorithm change. While I admire the cleverness and tenacity that it takes to solve a puzzle in a surprising way like this (which admittedly is most of what we do here), this code crosses a line for me in terms of how unintuitive it is. This is even after the good writeup that was added later on my request. I find it very hard to reason about what's going on, which is not a position I like being in as maintainer of the code :blush: 

I'm still willing to merge the code on the principle of "it works and passes all tests, so just shut up and accept your defeat", but it will need to bring real improvement. So far the evidence is not there.

htot:
I benchmarked on Ubuntu 19.10 using gcc version 9.2.1 20191008 (Ubuntu 9.2.1-9ubuntu2). Hashes and Short log  as a legend to the graphs (aqri/atom and aqrit/master last) :

35109ad | ab7a48b | 71a3509 | fb336ed | 67ee3fd | 0a69845 | 15e59e66 | 4cfed1c52
-- | -- | -- | -- | -- | -- | -- | --
V0.3.0 | Improve SSSE3 and AVX2 decoding speed | Fix invalid input handling | Add AVX code path | SSSE3->AVX2 encoding optimization | SSSE3->AVX2, NEON32 decoding optimization | aqrit/atom “add SSE2 dec_loop.c” | aqrit/master “comment SSSE3 dec_loop.c”

Decode benchmarks on the left, encode on the right:

afbeelding

On the encoding sidenot much change over time.
When decoding Plain saw a large improvement from 67ee3fd, AVX and SSE variants reach the current state at 0a69845.

afbeelding

On NUC (Atom Baytrail Silvermont) encoding improved with 67ee3fd.
When decoding SSSE3 reached its peak at ab7a48b, remained there until 67ee3fd. Then with 0a69845 AVX and SSE decoding regressed hard, so much even that it has the same speed as Plain.

On Atom Silvermont 67ee3fd has optimum speed for both encode and decode.

Also aqrits work neither speeds up, nor hurts performance on i7/atom architectures.

aklomp:
Thanks for the benchmarking. A few thoughts:

- On the i7, I find it strange that the AVX2 codec is not significantly faster than the others. It's definitely much faster on my 4th gen i5. Perhaps a bug in the benchmark?
- Decoding speeds on the NUC seem to have suffered over time. Maybe we can improve the platform check to include the processor. Then again, if you want the fastest decoding, don't use an Atom :wink:
- The SSE4 codecs have been simple aliases for SSSE3 since [67ee3fd](https://github.com/aklomp/base64/commit/67ee3fdbb154ad5cdb331f3ea40f5fa781613f81). Although I think an SSE4 codec could still bring some speedups by employing `_mm_testz_si128()`.
- In terms of performance, the low hanging fruit is probably to overhaul the scaffolding code. Get rid of Duff's Device, promote the `enc_loop` snippets to actual functions, etc. This is on my roadmap.
- I take it that the benchmarks support not merging the commits on this branch for the time being.

htot:
>  On the i7, I find it strange that the AVX2 codec is not significantly faster than the others. It's definitely much faster on my 4th gen i5. Perhaps a bug in the benchmark?

AVX2 decode has been much faster in the past. But after 0a69845 SSSE3 has caught up. May be there is another bottle neck (RAM is only DDR1600) .

>  Decoding speeds on the NUC seem to have suffered over time. 

Not exactly. Plain improved with 67ee3fd. SSSE3 regressed with 0a69845, the same commit  that brings improvement on i7. So, not a gradual degradation.

I found on SLM in particular in 64b mode, certain instructions incur a penalty. It seems to be something like that, a 50% improvement due to more clever algorithm combined by a 2x penalty.  Looking at Edison benchmarks here https://github.com/htot/base64 we see that indeed 64b is slower then 32b.

> Maybe we can improve the platform check to include the processor family. Then again, if you want the fastest decoding, don't use an Atom wink

My earlier benchmarks were on Edison (same core, no cache, 500MHz). I guess it's more of a perf/Watt game for me.

>  Get rid of Duff's Device, promote the `enc_loop` snippets to actual functions, etc. This is on my roadmap.

I like Duff's device. I took that idea from you and applied in crc32c. I managed to achieve same speed in C as in the original asm code.

>   I take it that the benchmarks support not merging the commits on this branch for the time being.

There seems to be no performance improvement on either platform.

htot:
Alright, I just had to try running 32b. On i7 there really is not much to see, but on SLM (Atom):

afbeelding

To make it build 32b I added CFLAGS+=-m32 on the make line but needed -fno-pie as well. And -m elf_i386 for ld.

Note the cases where running -m32-no-pie version is **faster**  then 64 bits. Nevertheless 67ee3fd remains faster than 0a69845 for SSSE3 decoding. So apparently there are 2 effects, possibly an instruction is used that is not HW implemented on Atom. And there is the 64b penalty I mentioned.

PR #44, opened 2017-10-21
Updated benchmarks on i7 and Edison

htot:
I also reran benchmarks at earlier commits to track improvement/degradation in coding/decoding speeds.

Strangely it looks like Edison SSE Decode has degraded by recent commits to the level of Plain. The same commit is creating an improvement on i7.

The exact commit causing this is 0a69845. It reduced SSE3 decode on Edison from 207 -> 146

Very strange.


aklomp:
Hey, sorry for the late response to this PR. Unfortunately I decided not to merge it, for the following reasons:

- As far as I can tell, the PR does not come with a way to regenerate these graphs from the data, or to generate additional graphs. Something like a Makefile would be nice.
- I'm not in favor of committing large binary blobs to this repo. Git doesn't like them, and it's not very friendly to users. The graphs should be in SVG format.
- The benchmarks themselves are, or will be, obsolete soon. I'm working on some substantial performance improvements, see #58 and #59.
- In my opinion, these images should be stored in `tests/benchmarks/images`, not in a new toplevel directory.

I've been thinking about how we store and display the benchmark results, and created issue #60 to track my thoughts. You're welcome to join that discussion!

htot:
The performance degradation on Atom (Silvermont) was the real news.

PR #45, opened 2017-10-22
Allow unity builds

myfreeweb:
Thanks a lot for the library! I made [a Golang wrapper](https://github.com/myfreeweb/go-base64-simd), building C code into a single object file [like this](https://github.com/myfreeweb/go-base64-simd/blob/5b6f2d8a68850717af4bf539e0dc378bd1ca88d1/c/combo.c). This patch resolves name conflicts, allowing that kind of build to succeed.

### DIFF
--- a/lib/arch/avx2/codec.c
+++ b/lib/arch/avx2/codec.c
@@ -8,12 +8,12 @@
 #if HAVE_AVX2
 #include <immintrin.h>
 
-#define CMPGT(s,n)	_mm256_cmpgt_epi8((s), _mm256_set1_epi8(n))
-#define CMPEQ(s,n)	_mm256_cmpeq_epi8((s), _mm256_set1_epi8(n))
-#define REPLACE(s,n)	_mm256_and_si256((s), _mm256_set1_epi8(n))
-#define RANGE(s,a,b)	_mm256_andnot_si256(CMPGT((s), (b)), CMPGT((s), (a) - 1))
+#define CMPGT256(s,n)	_mm256_cmpgt_epi8((s), _mm256_set1_epi8(n))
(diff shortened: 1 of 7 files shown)


aklomp:
Hi, it's really cool to see someone building wrappers for this lib in a different language. Kudos for that! 

However, I don't think your approach is portable. You [hard-code](https://github.com/myfreeweb/go-base64-simd/blob/master/c/config.h) all `HAVE_something` #define's to 1, and [compile the lib](https://github.com/myfreeweb/go-base64-simd/blob/master/base64/Makefile) with all SIMD architecture flags hard-coded to on. This means that effectively all code is compiled as AVX2, including the fallback code for older platforms. That won't end well, because the compiler might (will) decide to optimize to AVX2 level.

The proper way would be to define `AVX2_CFLAGS`, `SSSE3_CFLAGS`, etc individually, so that the base64 Makefile can selectively enable the compiler flags on a per-file basis. The SSSE3 codec is made only with `SSSE3_CFLAGS`, and so on. You can see how this breaks your concept of concatenating all code into one file and compiling it with one set of compiler flags.

myfreeweb:
The C code is not compiled on the target machine, the object file is downloaded pre-compiled! The Makefile is only run by me, in development. The idea was to avoid `cgo` entirely (it is used for now, but might be replaced by Go assembly).

I have tested it on a machine without AVX2, as you can see in the README :)

myfreeweb:
I guess Go might be able to link multiple object files though… Yeah, the separate build for safety makes sense.

aklomp:
There's a chance that it works if you're lucky, but that's not something to rely on. You really should be compiling each file individually with just the architecture-specific machine flags for that codec.

Do you mind if we close this PR then, if concatenating the files is out?

PR #57, opened 2019-11-21
Rebuild if any of the CFLAGS changed

miniupnp:


aklomp:
Thanks for the contribution. Not sure I'm entirely on board with it yet. So this would cause the library to automatically recompile when the compiler flags are changed in the environment? I can see how that would be useful in some specialized scenarios, like when building the library in a multi-target environment, but it seems to me like the outermost build system would need to handle that scenario. The normal workflow here would be to run `make clean` before recompiling.

Could you please provide a short rationale for this change?

miniupnp:
Well it is made so "make clean" is not necessary anymore before building.
It is indeed useless if you always start from a fresh state.

aklomp:
I intend to close this PR without merging it. I'm still not really on board with the change, as my earlier comment clarifies, and there have also been a lot of changes recently in how this library can be built. In particular, the library can now be built with `CMake`.

PR #71, opened 2020-10-05
Add basic CMake support

opeik:
Lacks feature detection but that shouldn't be too much extra work.


opeik:
Duplicate of #7. Deleting.

PR #75, opened 2021-11-14
chore: move x64 CI to GHA

mayeut:
Travis CI does not offer free x64 CI for OpenSource projects anymore. Move to GitHub Actions for x64 tests.

Sample build is there: https://github.com/mayeut/base64/actions/runs/2323476527

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,45 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  makefile-test:
+    name: makefile-${{ matrix.runner }}-amd64-${{ matrix.compiler }} ${{ ((matrix.openmp == 1) && '+openmp') || '' }}
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: ["ubuntu-18.04"]
+        compiler: ["gcc", "clang"]
+        openmp: ["0", "1"]
+        include:
+          - runner: "macos-11"
+            compiler: "clang"
+            openmp: "0"
+    env:
+      OPENMP: ${{ matrix.openmp }}
+      OMP_NUM_THREADS: ${{ ((matrix.openmp == 1) && '2') || '0' }}
+      CC: ${{ matrix.compiler }}
+      OBJCOPY: ${{ (startsWith(matrix.runner, 'macos') && 'echo') || 'objcopy' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Run tests
+        run: ./test/ci/amd64.sh
+
+  cmake-test:
+    name: cmake-${{ matrix.runner }}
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
+name: Test
+
+on:
+  push:
+  pull_request:
```
aklomp, line 5: I think this can be an array?

```yaml
on: [push, pull_request]
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+            compiler: "clang"
+            openmp: "0"
+    env:
+      OPENMP: ${{ matrix.openmp }}
+      OMP_NUM_THREADS: ${{ ((matrix.openmp == 1) && '4') || '0' }}
+      CC: ${{ matrix.compiler }}
+      OBJCOPY: ${{ (startsWith(matrix.runner, 'macos') && 'echo') || 'objcopy' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
aklomp, line 28: This one's already at `v3`.

.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: ["ubuntu-20.04"]
+        compiler: ["gcc", "clang"]
+        openmp: ["0", "1"]
+        include:
+          - runner: "macos-11"
+            compiler: "clang"
+            openmp: "0"
```
aklomp, line 20: I believe GitHub also has Windows runners available? Well, those could be added later, no action needed.

test/ci/amd64.sh
```diff
@@ -5,10 +5,17 @@ export SSSE3_CFLAGS=-mssse3
 export SSE41_CFLAGS=-msse4.1
 export SSE42_CFLAGS=-msse4.2
 export AVX_CFLAGS=-mavx
-export AVX2_CFLAGS=-mavx2
+# no AVX2 on GHA macOS
+if [ "${OBJCOPY:-}" != "echo" ]; then
```
aklomp, line 9: This is a bit of a hack. It's probably nicer to do OS detection than to rely on `OBJCOPY` being set to the magic string `"echo"`.

.github/workflows/test.yml
mayeut: done.

.github/workflows/test.yml
mayeut: done

.github/workflows/test.yml
mayeut: windows now handled with cmake.

test/ci/amd64.sh
mayeut: done

mayeut:
Tests with CMake have been added which showed an issue with macOS build.

PR #76, opened 2021-11-14
fix: alpine gcc does not define _INTEGRAL_MAX_BITS/__WORDSIZE

mayeut:
alpine gcc does not define `_INTEGRAL_MAX_BITS`/`__WORDSIZE`
This leads to `BASE64_WORDSIZE` being defined to `0`.
Fallback to `__LONG_WIDTH__` if it's defined or error out.

Builds on top of #75

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,30 @@
+name: Test
+
+on:
+  push:
+  pull_request:
+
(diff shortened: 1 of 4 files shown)


mayeut:
Now probably obsolete. Will check with CMake to get something better in.

PR #78, opened 2022-05-14
fix: alpine gcc does not define `_INTEGRAL_MAX_BITS`/`__WORDSIZE`

mayeut:
alpine gcc does not define `_INTEGRAL_MAX_BITS`/`__WORDSIZE`
This leads to `BASE64_WORDSIZE` being defined to 0.
Fallback to `__LONG_WIDTH__` if it's defined or error out.

Builds on top of https://github.com/aklomp/base64/pull/75

### DIFF
--- a/lib/env.h
+++ b/lib/env.h
@@ -43,10 +43,14 @@
 #endif
 
 // Detect word size:
-#ifdef _INTEGRAL_MAX_BITS
+#if defined(_INTEGRAL_MAX_BITS)
 #  define BASE64_WORDSIZE _INTEGRAL_MAX_BITS
-#else
+#elif defined(__WORDSIZE)
 #  define BASE64_WORDSIZE __WORDSIZE
+#elif defined (__LONG_WIDTH__)
+#  define BASE64_WORDSIZE __LONG_WIDTH__
+#else
+#  error BASE64_WORDSIZE_NOT_DEFINED
 #endif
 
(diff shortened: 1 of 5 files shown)


lib/env.h
```diff
@@ -43,10 +43,14 @@
...
-#ifdef _INTEGRAL_MAX_BITS
+#if defined(_INTEGRAL_MAX_BITS)
 #  define BASE64_WORDSIZE _INTEGRAL_MAX_BITS
-#else
+#elif defined(__WORDSIZE)
 #  define BASE64_WORDSIZE __WORDSIZE
+#elif defined (__LONG_WIDTH__)
+#  define BASE64_WORDSIZE __LONG_WIDTH__
+#else
+#  error BASE64_WORDSIZE_NOT_DEFINED
```
aklomp, line 53: This will print the string `BASE64_WORDSIZE_NOT_DEFINED`. Is that intentional or did you intend to define a macro of that name?

aklomp:
Nevermind, I merged #75 just now.

aklomp:
I should have asked this before I merged the code :smile_cat: but what is the added benefit of testing on Alpine over testing on Ubuntu? As long as both distros expose a similar basic Linux environment, the distro flavor should be irrelevant. I'm worried that adding Alpine has made CI more complex without pulling its own weight in terms of added coverage.

PR #79, opened 2022-05-14
Some fixes to cmake files

htot:
Hi,

Here are some fixes after the recent pull of cmake support.
- when enabling OpenMP also link against openmp lib, otherwise still not enabled
- not only build test tools but also install them (this is needed for Yocto recipe to pick them up and package)
- also build base64 command line encoder
- shutup warnings on recent versions of cmake (optional, @burn care to review this?)
- updated edison x86_64 benchmarks (optional)

My motherboard i7-4770 @ 3.4 GHz DDR1600 died, I won't be able to update these. I could provide a set on i7-10700 CPU @ 2.90GHz DDR4 3200?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,12 @@
 #     http://creativecommons.org/publicdomain/zero/1.0/
 #
 ########################################################################
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.10.2)
+
+# new dependent option syntax. We are already compliant
+if (POLICY CMP0127)
+    cmake_policy(SET CMP0127 NEW) 
+endif()
 
 project(base64 LANGUAGES C VERSION 0.4.0)
 
@@ -42,7 +47,7 @@ cmake_dependent_option(BASE64_WITH_OpenMP "use OpenMP" OFF "OpenMP_FOUND" OFF)
 add_feature_info("OpenMP codec" BASE64_WITH_OpenMP "spreads codec work accross multiple threads")
 cmake_dependent_option(BASE64_REGENERATE_TABLES "regenerate the codec tables" OFF "NOT CMAKE_CROSSCOMPILING" OFF)
 
-set(_IS_X86 "\"${_TARGET_ARCH}\" STREQUAL \"x86\" OR \"${_TARGET_ARCH}\" STREQUAL \"x64\"")
+set(_IS_X86 "_TARGET_ARCH STREQUAL \"x86\" OR _TARGET_ARCH STREQUAL \"x64\"")
 cmake_dependent_option(BASE64_WITH_SSSE3 "add SSSE 3 codepath" ON ${_IS_X86} OFF)
 add_feature_info(SSSE3 BASE64_WITH_SSSE3 "add SSSE 3 codepath")
 cmake_dependent_option(BASE64_WITH_SSE41 "add SSE 4.1 codepath" ON ${_IS_X86} OFF)
@@ -54,11 +59,11 @@ add_feature_info(AVX BASE64_WITH_AVX "add AVX codepath")
 cmake_dependent_option(BASE64_WITH_AVX2 "add AVX 2 codepath" ON ${_IS_X86} OFF)
 add_feature_info(AVX2 BASE64_WITH_AVX2 "add AVX2 codepath")
 
-set(_IS_ARM "\"${_TARGET_ARCH}\" STREQUAL \"arm\"")
+set(_IS_ARM "_TARGET_ARCH STREQUAL \"arm\"")
 cmake_dependent_option(BASE64_WITH_NEON32 "add NEON32 codepath" OFF ${_IS_ARM} OFF)
 add_feature_info(NEON32 BASE64_WITH_NEON32 "add NEON32 codepath")
 
-set(_IS_ARM64 "\"${_TARGET_ARCH}\" STREQUAL \"arm64\"")
+set(_IS_ARM64 "_TARGET_ARCH STREQUAL \"arm64\"")
 cmake_dependent_option(BASE64_WITH_NEON64 "add NEON64 codepath" ON ${_IS_ARM64} OFF)
 add_feature_info(NEON64 BASE64_WITH_NEON64 "add NEON64 codepath")
 
@@ -181,7 +186,6 @@ define_SIMD_compile_flags()
 if (_TARGET_ARCH STREQUAL "x86" OR _TARGET_ARCH STREQUAL "x64")
     macro(configure_codec _TYPE)
         if (BASE64_WITH_${_TYPE})
-            message(STATUS "Add codec: lib/arch/${_DIR}/codec.c")
             string(TOLOWER "${_TYPE}" _DIR)
             set_source_files_properties("lib/arch/${_DIR}/codec.c" PROPERTIES
                 COMPILE_FLAGS "${COMPILE_FLAGS_${_TYPE}}"
@@ -220,11 +224,7 @@ configure_file("${CMAKE_CURRENT_LIST_DIR}/cmake/config.h.in" "${CMAKE_CURRENT_BI
 ########################################################################
 # OpenMP Settings
 if (BASE64_WITH_OpenMP)
-    target_compile_options(base64
-        PRIVATE
-            ${OpenMP_C_FLAGS}
-    )
-    target_link_libraries(base64 OpenMP::OpenMP_C)
+    target_link_libraries(base64 PRIVATE OpenMP::OpenMP_C)
 endif()
 
 ########################################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,6 +224,7 @@ if (BASE64_WITH_OpenMP)
         PRIVATE
             ${OpenMP_C_FLAGS}
     )
+    target_link_libraries(base64 OpenMP::OpenMP_C)
 endif()
 
 ########################################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,6 +234,10 @@ if (BASE64_BUILD_TESTS)
 endif()
 
 ########################################################################
+# base64
+add_subdirectory(bin)
+
+########################################################################
 # cmake install
 install(DIRECTORY include/ TYPE INCLUDE)
 install(TARGETS base64 EXPORT base64-targets)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ set_package_properties(OpenMP PROPERTIES
 ########################################################################
 # Compilation options
 option(BASE64_WERROR "Treat warnings as error" ON)
-option(BASE64_BUILD_TESTS "add test projects" ON)
+option(BASE64_BUILD_TESTS "add test projects" OFF)
 cmake_dependent_option(BASE64_WITH_OpenMP "use OpenMP" OFF "OpenMP_FOUND" OFF)
 add_feature_info("OpenMP codec" BASE64_WITH_OpenMP "spreads codec work accross multiple threads")
 cmake_dependent_option(BASE64_REGENERATE_TABLES "regenerate the codec tables" OFF "NOT CMAKE_CROSSCOMPILING" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,13 +14,14 @@ cmake_minimum_required(VERSION 3.10.2)
 
 # new dependent option syntax. We are already compliant
 if (POLICY CMP0127)
-    cmake_policy(SET CMP0127 NEW) 
+    cmake_policy(SET CMP0127 NEW)
 endif()
 
 project(base64 LANGUAGES C VERSION 0.4.0)
 
 include(GNUInstallDirs)
 include(CMakeDependentOption)
+include(CheckIncludeFile)
 include(FeatureSummary)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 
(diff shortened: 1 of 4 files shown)


test/CMakeLists.txt
```diff
@@ -16,13 +16,14 @@ function(add_base64_test TEST_NAME)
...
-    
+
     add_executable(${TEST_NAME} ${SRC_FILES})
     target_link_libraries(${TEST_NAME} PRIVATE base64)
-    
+
     add_test(NAME ${TEST_NAME}
         COMMAND ${TEST_NAME}
     )
+    install(TARGETS ${TEST_NAME} DESTINATION bin)
```
BurningEnlightenment, line 26: What is the use case for installing test executables?

CMakeLists.txt
```diff
@@ -224,6 +229,7 @@ if (BASE64_WITH_OpenMP)
         PRIVATE
             ${OpenMP_C_FLAGS}
     )
+    target_link_libraries(base64 OpenMP::OpenMP_C)
```
BurningEnlightenment, line 232: This is incompatible with our current minimum supported CMake version which is `3.1`. Personally I have no problems raising it, but I think we should make a concious decision here based on what CMake versions are available in widely used distributions like RHEL, ubuntu LTS, etc.
cc: @madebr

BurningEnlightenment (changes requested):

BurningEnlightenment:
> shutup warnings on recent versions of cmake

These stem from CMake behaviour changes. The given changes in `CMP0060`, `CMP0065`, `CMP0082`, `CMP0127` shouldn't affect us in a bad way. Whereas `CMP0060` is probably caused by the OpenMP target provided by newer Versions of cmake

BurningEnlightenment:
Sorry, for stringing you by, but I'm currently occupied with a university project. I think I might be able to further review this at the weekend. W.r.t. requiring a newer cmake version I did some research a few months a go:
> I think it should be considered to require an even newer version as a similar discussion on llvm-dev last year [[1]] revealed that all current LTS platforms support at least 3.6.1 (or 3.13.4 if one is willing to sacrifice RHEL 6 and Debian 9 support).

So from my POV 3.13.4 would be fine. However, I really would like to hear from other cmake users first 😅 like @madebr, @dotnwat or @vglavnyy 
 
[1]: https://lists.llvm.org/pipermail/llvm-dev/2020-April/140578.html

CMakeLists.txt
madebr: Minimum CMake version providing `OpenMP::OpenMP_C` is 3.9:
https://cmake.org/cmake/help/v3.9/module/FindOpenMP.html
This one was released in [July 2017](https://cmake.org/pipermail/cmake/2017-July/065814.html)

3.9 is available in Ubuntu 18.04 LTS (which will be EOL next year): https://launchpad.net/ubuntu/bionic/+source/cmake

Personally, I'm completely fine with bumping `cmake_minimum_required`.

small suggestion:
```suggestion
    target_link_libraries(base64 PRIVATE OpenMP::OpenMP_C)
```

bin/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+# executable has same name as lib, see https://gist.github.com/jlgerber/eafc4ee2b9954e27dd2bb009496b1b03
+add_executable(base64-bin base64.c)
+target_link_libraries(base64-bin PRIVATE base64)
+install(TARGETS base64-bin DESTINATION bin)
```
madebr, line 4: Let's use the cache variables, initialized by `GNUInstallDirs`.
```suggestion
install(TARGETS base64-bin RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}")
```

test/CMakeLists.txt
madebr, line 26:
```suggestion
    install(TARGETS ${TEST_NAME} DESTINATION "${CMAKE_INSTALL_BINDIR}")
```

CMakeLists.txt
```diff
@@ -14,6 +14,11 @@ cmake_minimum_required(VERSION 3.1)
 
 project(base64 LANGUAGES C VERSION 0.4.0)
 
+cmake_policy(SET CMP0082 NEW)
+cmake_policy(SET CMP0060 NEW)
+cmake_policy(SET CMP0065 NEW)
+cmake_policy(SET CMP0127 NEW)
```
madebr, line 20: I think you need to do the following here:
```cmake
if(POLICY CMP0127)
```
Policy [`CMP0127`](https://cmake.org/cmake/help/latest/policy/CMP0127.html) is quire new.

CMakeLists.txt
madebr: Using [this github action](https://github.com/marketplace/actions/actions-setup-cmake), you can enforce a specific cmake version to ensure compatibility with older versions.

dotnwat:
> So from my POV 3.13.4 would be fine. However, I really would like to hear from other cmake users first 😅 like @madebr, @dotnwat or @vglavnyy

We should be fine with 3.13+, thanks for asking. But FWIW, Ubuntu Bionic LTS still has standard support for another year and that has 3.10.

CMakeLists.txt
```diff
@@ -29,6 +30,9 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 include(TargetArch)
 detect_target_architecture(_TARGET_ARCH)
 
+CHECK_INCLUDE_FILE("<getopt.h>" HAVE_GETOPT_H)
+cmake_dependent_option(BASE64_BUILD_CLI "Build the cli for encoding and decoding" ON "HAVE_GETOPT_H" ON)
```
BurningEnlightenment, line 34:
```suggestion
cmake_dependent_option(BASE64_BUILD_CLI "Build the cli for encoding and decoding" ON "HAVE_GETOPT_H" OFF)
```
If `<getopt.h>` is not available (like on windows), we can't build th CLI. Therefore we need to force it off.

However, the include check is borked (see my other comment).

CMakeLists.txt
```diff
@@ -29,6 +30,9 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 include(TargetArch)
 detect_target_architecture(_TARGET_ARCH)
 
+CHECK_INCLUDE_FILE("<getopt.h>" HAVE_GETOPT_H)
```
BurningEnlightenment, line 33:
```suggestion
CHECK_INCLUDE_FILE(getopt.h HAVE_GETOPT_H)
```

I misremembered the check include syntax, sorry 😅

BurningEnlightenment (changes requested):

test/CMakeLists.txt
BurningEnlightenment: I have no problem with this, because the tests are not installed when test creation is disabled.

CMakeLists.txt
madebr: It's just esthetics, but you can lower case the function. (CMake is case insensitive for functions/macros, but case sensitive for variables)

BurningEnlightenment: yeah, I know. However, I've been lazy and copy pasted the function name…

madebr (approved):

BurningEnlightenment (approved):
@aklomp I think this is ready to be merged :shipit: 

(I think a sane quality gate going forward would be two approvals from contributors who made substantial cmake contributions)

CMakeLists.txt
```diff
@@ -235,12 +239,23 @@ endif()
...
 # base64
-add_subdirectory(bin)
+if (BASE64_BUILD_CLI)
+    add_executable(base64-bin 
+        bin/base64.c
+    )
+    target_link_libraries(base64-bin PRIVATE base64)
+    set_target_properties(base64-bin PROPERTIES 
+        OUTPUT_NAME base64
+    )
```
aklomp, line 249: Trailing whitespace, also on lines 243 and 247.

aklomp (commented):
This commit really needs some more explanation in the commit body about why the tests are being disabled. Without context, I have no clue why this change was made, and my inclination is to reject it.

CMakeLists.txt
aklomp: This commit is lacking a lot of context. Which warnings are these? Why should they need to be "shut up"? May I suggest adding a bit of explanation to the commit body as to why this change is necessary?

CMakeLists.txt
```diff
@@ -10,14 +10,14 @@
 #     http://creativecommons.org/publicdomain/zero/1.0/
 #
 ########################################################################
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.10.2)
```
aklomp, line 13: For this commit too, it would be nice to have some context about _why_ we require this newer version.

CMakeLists.txt
aklomp: Nevermind I guess, this change is removed again by `386bfb`, so I will remove this commit during the rebase.

CMakeLists.txt
BurningEnlightenment: Have you heard of https://editorconfig.org/ ? Most editors support it and it has an [option] instructing editors to squelch those pesky trailing spaces automatically.

[option]: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#trim_trailing_whitespace

BurningEnlightenment:
> This commit really needs some more explanation in the commit body about why the tests are being disabled. Without context, I have no clue why this change was made, and my inclination is to reject it.

b3c29792f6cc58f2d006f3b5ada5a41626eb31bc contains install commands for all test binaries, because yocto (the package manager used by htot) won't pick them up which in turn is desirable if you want to verify that your cross compiled binaries work. However, in the _general_ case library consumers don't need the test binaries at all i.e. they are bloat and not worth to be compiled or installed. So I thought we should make the majority use case the default and put the configuration burden on the developing minority. Alternatively we can add another config option which controls whether the test binaries will be installed or not and turn that off by default instead.

htot:
> Edit: in reference to [8abb204a](https://github.com/aklomp/base64/pull/79/commits/8abb204a947ef8b46e1464fa82f655f02f221a3b)
> 
> This commit really needs some more explanation in the commit body about why the tests are being disabled. Without context, I have no clue why this change was made, and my inclination is to reject it.

I believe as a default, the idea is only the library is built. But with the cmake gui you can change these defaults during configure (as well as turn on openmp, turn off sse4 etc).

For me, the Yocto recipe overrides all defaults and builds everything, so the defaults don't matter. And then everything is packaged into it's respective deb package (-dev, -dbg, -tools etc.).

CMakeLists.txt
htot: Yes, it was discussed above. But that never made it into the commit message.

BurningEnlightenment: I would propose the following commit message:
```
Bump the required cmake version to `3.10.2`

CMake `3.9` introduced the library target `OpenMP::OpenMP_C` which is way more portable than the old compiler flag based approach. Due to the required version bump we surveyed the available CMake versions on widely used OS distributions. With said survey data (see below) we chose `3.10.2` as the new required version which only excludes RHEL7.

https://lists.llvm.org/pipermail/llvm-dev/2020-April/140578.html
```

CMakeLists.txt
aklomp: @BurningEnlightenment Never heard of it until now. I'm an oldschool guy who runs `vim` in a terminal and exclusively uses `git` through the command line interface, maybe that's why :)

I'm not a huge fan of things like modelines that mess with people's editor settings, and it looks like this might fall in that category. But as maintainer of this project, I do insist on adherence to a certain formatting, such as tabs for indent, no trailing whitespace, and so on. If an `.editorconfig` can politely warn the user about things that run counter to the intended formatting, then I do see the value of adding that file. Until then, I'm also fine with doing a manual style check before merging.

I've also considered setting up a style checker in CI (something like `uncrustify` or `clang-format`), but I think it would be a bit annoying, because people would need to run the tool locally to make sure that their code passes CI.

@htot No worries, that's what code review is for :)

CMakeLists.txt
aklomp: That's a great commit body, thanks.

aklomp:
@htot Thanks for clarifying your use-case a bit. I'm slightly familiar with Yocto and have run into it in the past when trying to recompile some software for an embedded board. Man, what a monster, I much prefer Buildroot :)

CMakeLists.txt
BurningEnlightenment: Oh, I'm also a CLI person, but I seldomly use vim because that is/was not so well supported on Windows.

I'm not familiar with vim `modelines`, but an `.editorconfig` can be seen as a very basic but universal `.clang-format`. It instructs editors to (temporarely) apply the configured formatting settings (only) within the containing directory. It only supports the options `indent_style`, `tab_width`, `end_of_line` (style), `charset`, `trim_trailing_whitespace`, `insert_final_newline`, but as opposed to `.clang-format` it is more widely applicable and natively supported by editors. People who get disturbed by this usually have disabled/not installed editorconfig support globally.

Adding a `.clang-format` is a good idea, too. Quite a few editors have plugins/integrations for enforcing it on save (even [vim]), so chances are that people wouldn't need to manually run `clang-format`. However, `clang-format` can't format CMake files.

And just for the record there is an [editorconfig plugin] for vim, too 😉 

[editorconfig plugin]: https://github.com/editorconfig/editorconfig-vim#readme
[vim]: https://github.com/rhysd/vim-clang-format

aklomp: @BurningEnlightenment Thanks for explaining, I created #81 to add an `.editorconfig` file.

htot:
I think I resolved all comments now and set Kate to always removing training white spaces. I squashed the 2 commits and added suggestions to commit messages. Thanks all for review.

aklomp:
Thanks @htot. I'll merge this patchset, but I do think it's a bit messy:

- The changes to the benchmarks do not fit with the theme of this issue and are sort of sneaked in.
- This patchset first introduces some code (the workaround to install an executable with the same name as a library) and then removes it in a subsequent commit. I'm a fan of making commits atomic and composable: they only add or remove something, and do so in isolation without depending on other commits in the patch set. Anwyay, this project has no submission guidelines yet, so it's not a big deal.

Thanks everyone for contributing!

PR #82, opened 2022-06-04
Benchmarks: fix detection of supported codecs on x86

htot:
This fixes codec_supported as used by `benchmarks`. `benchmarks` forces all codecs built-in at compile time. It does run a test on a test string but this can cause an exception with certain codecs. To this end we factor out cpu capability detection from lib/codec_choose, introduce an additional flag codec flag BASE64_CHECK_SUPPORT, when set this performs cpu capability detection when forced prior to running the test and prevents running the codec if not supported. This ways existing code running the library is not affected.

Signed-off-by: Ferry Toth <ftoth@exalondelft.nl>

### DIFF
--- a/include/libbase64.h
+++ b/include/libbase64.h
@@ -53,6 +53,7 @@ extern "C" {
 #define BASE64_FORCE_SSE41	(1 << 5)
 #define BASE64_FORCE_SSE42	(1 << 6)
 #define BASE64_FORCE_AVX	(1 << 7)
+#define BASE64_CHECK_SUPPORT	(1 << 15)
 
 struct base64_state {
 	int eof;
--- a/lib/codec_choose.c
+++ b/lib/codec_choose.c
@@ -84,57 +84,84 @@ BASE64_CODEC_FUNCS(sse41)
 BASE64_CODEC_FUNCS(sse42)
 BASE64_CODEC_FUNCS(avx)
 
+static bool avx2_supported(void);
+static bool avx_supported(void);
+static bool sse42_supported(void);
+static bool sse41_supported(void);
+static bool ssse3_supported(void);
+
 static bool
 codec_choose_forced (struct codec *codec, int flags)
 {
-	// If the user wants to use a certain codec,
-	// always allow it, even if the codec is a no-op.
+	bool check;
+	// If the user wants to use a certain codec, always allow it, 
+	// even if the codec is a no-op, except when BASE64_CHECK_SUPPORT
+	// is set.
 	// For testing purposes.
 
+	check = flags & BASE64_CHECK_SUPPORT;
+	flags = flags & ~BASE64_CHECK_SUPPORT;
 	if (!(flags & 0xFF)) {
 		return false;
 	}
 	if (flags & BASE64_FORCE_AVX2) {
-		codec->enc = base64_stream_encode_avx2;
-		codec->dec = base64_stream_decode_avx2;
-		return true;
+		if (!check || avx2_supported()) {
+			codec->enc = base64_stream_encode_avx2;
+			codec->dec = base64_stream_decode_avx2;
+		} else {
+			codec->enc = NULL;
+			codec->dec = NULL;
+		}
 	}
 	if (flags & BASE64_FORCE_NEON32) {
 		codec->enc = base64_stream_encode_neon32;
 		codec->dec = base64_stream_decode_neon32;
-		return true;
 	}
 	if (flags & BASE64_FORCE_NEON64) {
 		codec->enc = base64_stream_encode_neon64;
 		codec->dec = base64_stream_decode_neon64;
-		return true;
 	}
 	if (flags & BASE64_FORCE_PLAIN) {
 		codec->enc = base64_stream_encode_plain;
 		codec->dec = base64_stream_decode_plain;
-		return true;
 	}
 	if (flags & BASE64_FORCE_SSSE3) {
-		codec->enc = base64_stream_encode_ssse3;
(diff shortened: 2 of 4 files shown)


lib/codec_choose.c
```diff
@@ -195,70 +220,179 @@ codec_choose_x86 (struct codec *codec)
...
-	return false;
+static bool
+codec_choose_x86 (struct codec *codec)
+{
+    if(avx2_supported()) {
+	    codec->enc = base64_stream_encode_avx2;
+	    codec->dec = base64_stream_decode_avx2;
+	    return true;
+    };
+    if( ssse3_supported()) {
```
htot, line 373: Oops, copy/paste error, should be avx_supported()

lib/codec_choose.c
```diff
@@ -195,70 +220,179 @@ codec_choose_x86 (struct codec *codec)
...
-				#endif
-				#if HAVE_AVX
+			}
+		}
+	}
+
+	return false;
+}
+#else
+static bool avx2_supported(void) {return false}
```
mayeut, line 236:
```suggestion
static bool avx2_supported(void) {return false;}
```

lib/codec_choose.c
```diff
@@ -195,70 +220,179 @@ codec_choose_x86 (struct codec *codec)
...
+	unsigned int eax, ebx = 0, ecx = 0, edx;
+	unsigned int max_level;
+
+#ifdef _MSC_VER
+	int info[4];
+	__cpuidex(info, 0, 0);
+	max_level = info[0];
+#else
+	max_level = __get_cpuid_max(0, NULL);
+#endif
```
aklomp, line 321: This code is duplicated and should be split out into a separate function.

lib/codec_choose.c
```diff
@@ -195,70 +220,179 @@ codec_choose_x86 (struct codec *codec)
...
-			codec->enc = base64_stream_encode_sse42;
-			codec->dec = base64_stream_decode_sse42;
 			return true;
 		}
 	}
-	#endif
+	return false;
+}
+#else
+static bool sse42_supported(void) {return false;}
```
aklomp, line 306: I don't like these tiny vestigial functions. They don't even follow the regular formatting of functions. My suggestion would be something like this instead:

```c
 static bool sse42_supported (void)
{
#if HAVE_SSE42
	...
#else
	return false;
#endif
}
```

lib/codec_choose.c
```diff
@@ -84,57 +84,84 @@ BASE64_CODEC_FUNCS(sse41)
 BASE64_CODEC_FUNCS(sse42)
 BASE64_CODEC_FUNCS(avx)
 
+static bool avx2_supported(void);
+static bool avx_supported(void);
+static bool sse42_supported(void);
+static bool sse41_supported(void);
+static bool ssse3_supported(void);
```
aklomp, line 91: These forward declarations can be elided by defining the functions before first use.

lib/lib.c
```diff
@@ -143,6 +144,14 @@ base64_decode
 {
 	int ret;
 	struct base64_state state;
+	bool check;
+
+	check = flags & BASE64_CHECK_SUPPORT;
+	if (check) {
```
aklomp, line 150: Bit verbose?

```c
if (flags & BASE64_CHECK_SUPPORT) {
```

aklomp: Also elides the inclusion of `<stdbool.h>`.

lib/codec_choose.c
aklomp: Also, the code style dictates a space after `if`:

```c
if (avx_supported()) {
```

But I can fix that in the rebase.

lib/codec_choose.c
aklomp:
```c
static inline unsigned int x86_cpuid_max_level (void)
{
#ifdef BASE64_X86_SIMD
#ifdef _MSC_VER
	int info[4U];

	__cpuidex(info, 0, 0);
	return info[0U];
#else
	return __get_cpuid_max(0, NULL);
#endif
#else
	return 0U;
#endif
}
```

include/libbase64.h
```diff
@@ -53,6 +53,7 @@ extern "C" {
 #define BASE64_FORCE_SSE41	(1 << 5)
 #define BASE64_FORCE_SSE42	(1 << 6)
 #define BASE64_FORCE_AVX	(1 << 7)
+#define BASE64_CHECK_SUPPORT	(1 << 15)
```
aklomp, line 56: This flag is not documented in `README.md`. I also don't understand why the bitmasks make a numerical jump.

aklomp: I'm not sure if the flag makes sense. I would expect that even when the user forces a codec, we use runtime feature detection to check whether the codec is actually available. I can't see the use case for wanting to force an AVX2 codec on some low power machine and expecting anything sane to happen. If anything, the sane thing would be to return `false`.

aklomp:
Thnaks @htot, this is an interesting pull request. I like where this is going and I think it's a good idea to factor out the x86 feature detection bits. I have some ideas about how to enhance this concept and I may come up with a counter-proposal of my own.

include/libbase64.h
htot: True, but that is what the comment said (force even if there is no CPU support) and I was not sure if I would be disturbing any test case by changing the default behavior. Then the bitmask is misused (not to add a codec but to change behavior), so I thought to leave room for future codecs.

But from my pov it would be best to not add the bitmask at all. So, if you agree...

htot:
Sure, I'll wait to see what you come up with before attempting a v2.

PR #84, opened 2022-06-05
fix(ci): cmake tests are not run on GHA

mayeut:
Following #79, tests were not run anymore with cmake builds (as tests were not built by default after that PR).
This PR re-enables testing cmake builds, adds `--no-tests=error` flag to `ctest` (would have caught the "no tests" == OK issue).
This also adds `-Werror=dev` to `cmake` configure step to check and enforce the effectiveness of #79 regarding the multiple warnings that were output by cmake before that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,11 +38,17 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: CMake Configure
-        run: cmake -B out -DCMAKE_BUILD_TYPE=Release ${{ runner.os == 'macOS' && '-DBASE64_WITH_AVX2=OFF' || '' }}
+        run: >
+          cmake
+          -B out
+          -Werror=dev
+          -DBASE64_BUILD_TESTS=ON
+          ${{ runner.os != 'Windows' && '-DCMAKE_BUILD_TYPE=Release' || '' }}
+          ${{ runner.os == 'macOS' && '-DBASE64_WITH_AVX2=OFF' || '' }}
       - name: CMake Build
         run: cmake --build out --config Release --verbose
       - name: CTest
-        run: ctest --test-dir out -VV --build-config Release
+        run: ctest --no-tests=error --test-dir out -VV --build-config Release
 
   alpine-makefile-test:
     name: makefile-alpine-amd64-gcc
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -72,9 +72,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: CMake Configure
-        run: cmake -B out -DCMAKE_BUILD_TYPE=Release ${{ runner.os == 'macOS' && '-DBASE64_WITH_AVX2=OFF' || '' }}
+        run: cmake -B out -Werror=dev -DBASE64_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release
```
aklomp, line 75: Why did the conditional statement disappear?

aklomp: Ah, it was moved to line 41. Nevermind.

.github/workflows/test.yml
```diff
@@ -38,11 +38,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: CMake Configure
-        run: cmake -B out -DCMAKE_BUILD_TYPE=Release ${{ runner.os == 'macOS' && '-DBASE64_WITH_AVX2=OFF' || '' }}
+        run: cmake -B out -Werror=dev -DBASE64_BUILD_TESTS=ON ${{ runner.os != 'Windows' && '-DCMAKE_BUILD_TYPE=Release' || '' }} ${{ runner.os == 'macOS' && '-DBASE64_WITH_AVX2=OFF' || '' }}
```
aklomp, line 41: This command should be split as follows for readability:

```yaml
run: >
  cmake
  -B out
  -Werror=dev
  -DBASE64_BUILD_TESTS=ON
  ${{runner.os != 'Windows' && '-DCMAKE_BUILD_TYPE=Release' || ''}}
  ${{runner.os == 'macOS' && '-DBASE64_WITH_AVX2=OFF' || ''}}
```

mayeut: done

.github/workflows/test.yml
mayeut: alpine is always linux, no need for any conditional here.

PR #87, opened 2022-06-05
Update word size detection

mayeut:
The detection for word size was dependent on `__LONG_WIDTH__` for at least gcc Alpine Linux (probably other musl distros too).
This is not working with clang which does not define `__LONG_WIDTH__`.
Replace `__LONG_WIDTH__` by `__SIZE_WIDTH__`.
As this will not work as expected for x32 ABI, add a robust x32 detection first.

### DIFF
--- a/lib/env.h
+++ b/lib/env.h
@@ -43,12 +43,15 @@
 #endif
 
 // Detect word size:
-#if defined(_INTEGRAL_MAX_BITS)
+#if defined(__x86_64__) && defined(__ILP32__)
+// this is the way to detect x32 ABI per https://wiki.debian.org/X32Port
+#  define BASE64_WORDSIZE 64
+#elif defined(_INTEGRAL_MAX_BITS)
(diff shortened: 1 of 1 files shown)


lib/env.h
```diff
@@ -43,12 +43,15 @@
 #endif
 
 // Detect word size:
-#if defined(_INTEGRAL_MAX_BITS)
+#if defined(__x86_64__) && defined(__ILP32__)
+// this is the way to detect x32 ABI per https://wiki.debian.org/X32Port
+#  define BASE64_WORDSIZE 64
```
mayeut, line 48: I wanted to keep the old behavior for platforms other than x32 ABI but this check could be simplified in this specific instance to just:
```
#if defined(__x86_64__)
  // force word size on __x86_64__ 
  // on x32 ABI which defines __x86_64__ and uses 64-bit words, __SIZE_WIDTH__ == 32
  #  define BASE64_WORDSIZE 64
```

aklomp: Indeed, I'll fix it in the rebase.

PR #88, opened 2022-06-06
Add an editorconfig codifying the current coventions

### LINKED ISSUE #81: .editorconfig: add dotfile
Add an `.editorconfig` dotfile in the root of the project to help enforce a consistent code style. This idea was raised in the discussion for #79, as a way to avoid preventable mistakes such as trailing whitespace or using an inconsistent spacing style.

The `.editorconfig` format is documented at [editorconfig.org](https://editorconfig.org).

BurningEnlightenment:
Resolves #81


.editorconfig
```diff
@@ -0,0 +1,16 @@
+# https://EditorConfig.org
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+indent_style = tab
+tab_width = 4
```
aklomp, line 10: Actually I use 8-space tabs, because I consider it the standard for C code.

I think 4 or 8 shouldn't matter with cautious use of tabs (only at start of line and only for adding nesting levels), but I can't rule out that some spacing might get mangled.

I bring this up because adding this to the `.editorconfig` makes GitHub display the code with 4-space tabs.

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+indent_size = 4
```
aklomp, line 11: Missing directive:

```
end_of_line = lf

```

BurningEnlightenment: I left that one intentionally unset, because it would conflict with git's `autocrlf` feature on Windows. If we wanted `lf` line endings everywhere we would need to reconfigure git via `.gitattributes`. However, this will cause conflicts with existing clones on windows.

.editorconfig
BurningEnlightenment: Sorry, fixed.

.editorconfig
aklomp: Hm yes, the `.editorconfig` controls what the editor does in the user's local environment, and if that environment is Windows, then I can imagine that the user wants "the usual" line terminators. I can respect not messing with the user's preference in their own environment.

However, *for commits to this project* I would like to standardize on `\n` as the line ending. I guess the `.editorconfig` might not be the best place to encode that, and a better place would indeed be `.gitattributes`, or perhaps some sort of style check in CI.

aklomp: But isn't the `.editorconfig` file intended as a hint to the editor on how to save files that are intended to be *commits for this project*? In that case I think setting the end-of-line preference might be fair after all.

I think it shouldn't conflict with `autocrlf` because when Git checks out the files, it will change `\n` to `\r\n`, but then the editor will map them back to `\n`.

.editorconfig
mayeut: Given `indent_style = tab`, can't `tab_width = 4` just be omitted as it's just cosmetic & will not modify in anyway what will end up in the files ?
I'd rather use user default setting for this one (there's also a default per user GitHub setting for this).

aklomp:
There's a typo in the commit title (`coventions` -> `conventions`), mind if I fix it during the merge? Or would you consider updating the commit?

aklomp:
Merged after rebasing on top of the latest `master`.

Issue #96, opened 2022-07-17
clang build fails with inline ASM on NEON64 (Apple M1)


mscdex:
clang must not be allocating `l3` in a contiguous register? While building 3eab8e6ca57f4514dea4fd0ec435967e96371bbe, the compiler errors are:

```
In file included from ../../deps/base64/base64/lib/arch/neon64/codec.c:62:
../../deps/base64/base64/lib/arch/neon64/enc_loop.c:32:44: error: registers must be sequential
                "and  %[t3].16b, v14.16b,   %[n63].16b \n\t"
                                                         ^
<inline asm>:10:40: note: instantiated into assembly here
        tbl v12.16b, {v5.16b, v6.16b, v7.16b, v16.16b}, v3.16b
                                              ^
In file included from ../../deps/base64/base64/lib/arch/neon64/codec.c:62:
../../deps/base64/base64/lib/arch/neon64/enc_loop.c:32:44: error: unknown token in expression
                "and  %[t3].16b, v14.16b,   %[n63].16b \n\t"
                                                         ^
<inline asm>:10:48: note: instantiated into assembly here
        tbl v12.16b, {v5.16b, v6.16b, v7.16b, v16.16b}, v3.16b
                                                      ^
In file included from ../../deps/base64/base64/lib/arch/neon64/codec.c:62:
../../deps/base64/base64/lib/arch/neon64/enc_loop.c:32:44: error: invalid operand
                "and  %[t3].16b, v14.16b,   %[n63].16b \n\t"
                                                         ^
<inline asm>:10:48: note: instantiated into assembly here
        tbl v12.16b, {v5.16b, v6.16b, v7.16b, v16.16b}, v3.16b
                                                      ^
In file included from ../../deps/base64/base64/lib/arch/neon64/codec.c:62:
../../deps/base64/base64/lib/arch/neon64/enc_loop.c:35:75: error: registers must be sequential
                "tbl v12.16b, {%[l0].16b, %[l1].16b, %[l2].16b, %[l3].16b}, %[t0].16b \n\t"
                                                                                        ^
<inline asm>:11:40: note: instantiated into assembly here
        tbl v13.16b, {v5.16b, v6.16b, v7.16b, v16.16b}, v2.16b
                                              ^
In file included from ../../deps/base64/base64/lib/arch/neon64/codec.c:62:
../../deps/base64/base64/lib/arch/neon64/enc_loop.c:35:75: error: unknown token in expression
[...]

PR #97, opened 2022-07-18
NEON64: enc: no inline asm for clang -O0

mscdex:
This may not be the best fix, but it does allow compiling of debug builds with clang.

Also, as far as I can tell the NEON32 inline asm is not affected.

Fixes https://github.com/aklomp/base64/issues/96

### DIFF
--- a/lib/arch/neon64/codec.c
+++ b/lib/arch/neon64/codec.c
@@ -18,8 +18,12 @@
 #include <arm_neon.h>
 
 // Only enable inline assembly on supported compilers.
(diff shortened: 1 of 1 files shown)


aklomp:
Hm, as mentioned in #96, I think there might be an easier fix, which is to have `clang` always load the encoding table with the `vld1q_u8_x4` intrinsic. As far as I know, that intrinsic gets compiled correctly by `clang`, and will allocate four sequential registers. (GCC does not compile the intrinsic correctly and chooses four random registers.)

PR #105, opened 2022-09-30
Fix out-of-bounds access in test

jkrems:
When running the tests with address sanitizer enabled, it fails with the following error:

```
test/test_base64.c:198:40: runtime error: index 402 out of bounds for type 'char[400]'
    #0 0x7f322c00c749 in test_streaming test/test_base64.c:198:40
    #1 0x7f322c00c749 in test_one_codec test/test_base64.c:340:10
    #2 0x7f322c00c749 in main test/test_base64.c:361:11
```

I think adding this bounds check preserves the semantics of the test but I'm not super familiar with the codebase.

### DIFF
--- a/test/test_base64.c
+++ b/test/test_base64.c
@@ -195,9 +195,14 @@ test_streaming (int flags)
 
 		base64_stream_decode_init(&state, flags);
 		memset(enc, 0, 400);
-		while (inpos < 400 && base64_stream_decode(&state, &ref[inpos], (inpos + bs > reflen) ? reflen - inpos : bs, &enc[enclen], &partlen)) {
+		while (base64_stream_decode(&state, &ref[inpos], (inpos + bs > reflen) ? reflen - inpos : bs, &enc[enclen], &partlen)) {
 			enclen += partlen;
 			inpos += bs;
+
+			// Has the entire buffer been consumed?
+			if (inpos >= 400) {
+				break;
+			}
 		}
 		if (enclen != 256) {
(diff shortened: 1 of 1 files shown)


aklomp:
Thanks for the PR. Looks like a real issue, an out-of-bounds pointer is being dereferenced.

Your fix works, and will bail out the loop before the dereference happens. The following code might make the intent a bit clearer:

```c
while (base64_stream_decode(&state, &ref[inpos], (inpos + bs > reflen) ? reflen - inpos : bs, &enc[enclen], &partlen)) {
     enclen += partlen;
     inpos += bs;

    // Has the entire buffer been consumed?
    if (inpos >= 400) {
        break;
    }
}
```

But no need to change it.

As a side note, I'd be interested in running `asan` in CI. Did you find this issue with `gcc` or `clang`'s address sanitizer, or were you using another tool?

jkrems:
Updated to use the suggested pattern.

More precisely, this was found via `clang`'s `--copt=-fsanitize=bounds`. Needed to look up the exact setup because I was building via bazel. The error will show up in this projects setup with:

```
make -C test CFLAGS=-fsanitize=bounds
```

That is, it will print an error though exit code is still 0. I didn't dig more deeply than that. :)

aklomp:
Sorry for reopening and reclosing, I forgot to annotate the commit with `Resolves #105`. Made some stealthy force-pushes to `master` to fix. Nothing to see here...

PR #110, opened 2022-10-22
Fix i686 builds

mayeut:
Inline assembly builds are failing in 32-bit, only enable AVX/AVX2 inline assembly in 64-bit mode.
This was already done for SSE* codecs.
Add tests for 32-bit builds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,15 +28,26 @@ jobs:
         run: ./test/ci/test.sh
 
   cmake-test:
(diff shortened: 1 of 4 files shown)


aklomp:
Oops. I knew there were 32-bit SSSE3 machines, but didn't realize that there was such a thing as AVX/AVX2 in a 32-bit context. Nice catch.

Thanks for adding the `i686` CI tests, so that this oversight won't happen again.

Issue #111, opened 2022-10-23
bug: out-of-bounds read when using inline assembly code path


mayeut:
While upgrading base64 in a project I'm working on, valgrind started to report out-of-bounds read in the optimized encoding functions.
Those are now using the inline assembly code path instead of the intrinsics code path.

A draft PR with a reproducer will follow (I do not plan to fix this though).

aklomp:
This is probably due to choosing the wrong [cutoff value](https://github.com/aklomp/base64/blob/master/lib/arch/avx2/enc_loop_asm.c#L129) for the source length. I'll look into it later. Thanks for reporting.

aklomp:
Found the bug. It's in [this](https://github.com/aklomp/base64/blob/master/lib/arch/ssse3/enc_loop_asm.c#L137) calculation. Note how it differs (mistakenly) from the [equivalent calculation](https://github.com/aklomp/base64/blob/master/lib/arch/ssse3/enc_loop.c#L31) in the non-asm code.

When calculating the number of rounds that the SIMD encoder can complete, we need to adjust for the fact that rounds consume 12 bytes of inputs, but the read size is 16 bytes. That's why we should subtract four bytes from the input size to ensure that a 16-byte read will not extend past the input buffer.

I'll fix and merge this after I review your pull request. The changes to CI in the pull request should validate the fix.

aklomp:
The bug occurred because the first encoder to be converted to inline asm was the AVX2 encoder. That codec [treats the first round specially](https://github.com/aklomp/base64/blob/master/lib/arch/avx2/enc_loop.c#L42) by reading it at an offset of +0 bytes, while all following rounds are read at an offset of -4 bytes. This led me to believe that subtracting the four bytes only applied to that codec, and not to AVX and SSSE3, which is why I left it out from those codecs.

PR #112, opened 2022-10-23
chore(ci): run static/dynamic analysis

mayeut:
This PR updates the workflow to run some static / dynamic analysis before running all other jobs.
It updates the unit test in order to get a reproducer for #111

### DIFF
--- a/test/ci/analysis.sh
+++ b/test/ci/analysis.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+set -ve
+
+MACHINE=$(uname -m)
+export CC=gcc
+
+uname -a
+clang --version  # make analyse
+${CC} --version  # make -C test valgrind
+
+for USE_ASSEMBLY in 0 1; do
+	if [ "${MACHINE}" == "x86_64" ]; then
+		export SSSE3_CFLAGS="-mssse3 -DBASE64_SSSE3_USE_ASM=${USE_ASSEMBLY}"
+		export SSE41_CFLAGS="-msse4.1 -DBASE64_SSE41_USE_ASM=${USE_ASSEMBLY}"
+		export SSE42_CFLAGS="-msse4.2 -DBASE64_SSE42_USE_ASM=${USE_ASSEMBLY}"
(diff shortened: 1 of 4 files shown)


test/ci/analysis.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set -ve
+
+MACHINE=$(uname -m)
+uname -a
+gcc --version
+clang --version
```
aklomp, line 7: `${CC} --version`, no? The rest of the script is compiler agnostic.

test/ci/analysis.sh
```diff
@@ -0,0 +1,65 @@
...
+	# export AVX512_CFLAGS="-mavx512vl -mavx512vbmi"
+elif [ "${MACHINE}" == "aarch64" ]; then
+	export NEON64_CFLAGS="-march=armv8-a"
+elif [ "${MACHINE}" == "armv7l" ]; then
+	export NEON32_CFLAGS="-march=armv7-a -mfloat-abi=hard -mfpu=neon"
+fi
+make clean
+make
+make -C test valgrind
+echo "::endgroup::"
```
aklomp, line 65: There's a lot of repetition in this file. I'd be tempted to avoid the copypasting and use a shell loop to enable/disable inline asm:

```sh
for ASM_ENABLED in 0 1; do
    ...
done
```

test/test_base64.c
```diff
@@ -107,8 +108,24 @@ test_char_table (int flags)
...
+			if (src == NULL) {
+				printf(
+					"FAIL: encoding @ %d: allocation of %lu bytes failed\n",
+					i, (unsigned long)chrlen
+				);
+				fail = true;
+				continue;
+			}
+			memcpy(src, &chr[i], chrlen);
+		}
```
aklomp, line 123: Nice, and very useful.

test/ci/analysis.sh
mayeut: `clang` is used explicitly in `make analyze`
I changed `gcc` by `${CC}`

test/ci/analysis.sh
mayeut: yes indeed

PR #121, opened 2023-11-05
Fix AVX detection

hardfalcon:
The old/faulty code would try to use AVX/AVX2 if either the SSE bit or the AVX bit were set in XCR0, but did not check if both bits were set.

In most cases, this still worked, but on some machines, enabling linux kernel mitigations for the "gather data sampling" vulnerability results in only the SSE bit but not the AVX bit being set, thus resulting in an illegal instruction and crashing the application.

Fix this by checking that both bits are set.

Fixes: 4bbb590bd047e7d877bf9cc9218ee525ceca1a9b ("Proper check of CPU's AVX2 feature support (with MSVC support)")

### DIFF
--- a/lib/codec_choose.c
+++ b/lib/codec_choose.c
@@ -208,7 +208,7 @@ codec_choose_x86 (struct codec *codec)
 		if (ecx & bit_XSAVE_XRSTORE) {
 			uint64_t xcr_mask;
 			xcr_mask = _xgetbv(_XCR_XFEATURE_ENABLED_MASK);
-			if (xcr_mask & _XCR_XMM_AND_YMM_STATE_ENABLED_BY_OS) {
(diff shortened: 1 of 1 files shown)


hardfalcon:
By the way, this issue also affects node.js, and projects that use node.js, like electron or [lightdm-webkit2-greeter](https://github.com/anthraxx/linux-hardened/issues/85).

aklomp:
Merged, thanks for the contribution!

hardfalcon:
You're welcome, thanks for merging so quickly :)

aklomp:
@bnoordhuis Indeed, it's time. I created #122 and will get on it right now.

aklomp:
Alright, released v0.5.1.

bnoordhuis:
Excellent, thanks!

PR #123, opened 2023-11-13
Detect word size for Windows properly

zcbenz:
In Windows SDK the `_INTEGRAL_MAX_BITS` is always 64 even when building for x86_32 targets, so it should not be used for detecting word size when compiling for Windows.

Also `_INTEGRAL_MAX_BITS` means the max bits of integer, which does not necessarily equal to the word size, so it should be used as the last resort for detecting word size.

Currently compiling Node.js for 32bit Windows with clang breaks because of this.

### DIFF
--- a/lib/env.h
+++ b/lib/env.h
@@ -46,12 +46,18 @@
 #if defined (__x86_64__)
 // This also works for the x32 ABI, which has a 64-bit word size.
 #  define BASE64_WORDSIZE 64
-#elif defined (_INTEGRAL_MAX_BITS)
-#  define BASE64_WORDSIZE _INTEGRAL_MAX_BITS
 #elif defined (__WORDSIZE)
 #  define BASE64_WORDSIZE __WORDSIZE
 #elif defined (__SIZE_WIDTH__)
 #  define BASE64_WORDSIZE __SIZE_WIDTH__
+#elif defined (_WIN32)
+// _WIN32 is defined for all Windows targets but _WIN64 is only defined for
+// 64-bit targets.
+#  if defined (_WIN64)
+#    define BASE64_WORDSIZE 64
+#  else
+#    define BASE64_WORDSIZE 32
+#  endif
 #else
 #  error BASE64_WORDSIZE_NOT_DEFINED
 #endif


aklomp:
The use of the `_INTEGRAL_MAX_BITS` macro was introduced in [fcab175](https://github.com/aklomp/base64/commit/fcab175) (by @BurningEnlightenment ) as a way to detect word size on MSVC. This patch will only check that macro on non-Windows systems. Then what is the point of checking the macro? Can't it better be removed entirely if it's not checked on the platform it was introduced for?

zcbenz:
Thanks for the background, I have removed the usage of `_INTEGRAL_MAX_BITS` completely.

For compatibility with MSVC, I have added another way to detect word size, which is officially documented.

BurningEnlightenment:
Maybe I'm missing something, but wouldn't it be possible to add two "generic"/C99 branches based on `SIZE_MAX==0xffffffff` and `SIZE_MAX==0xffffffffffffffff`? These should cover the Windows case as there isn't something like x32 on Windows. And as a bonus would make the library a little bit more portable.

zcbenz:
The purpose of this macro is to conditionally enable code in compile time, I think `SIZE_MAX==0xffffffff` only works in runtime.

BurningEnlightenment:
`SIZE_MAX` is a C99 preprocessor macro, i.e. it is guaranteed to work at compile time / in preprocessor expressions (see [cppreference] and this simple [godbolt-example]). The part I'm unsure about is whether `SIZE_MAX` (resp. `sizeof(size_t)`) is a good predictor for the wordsize on most machines and therefore whether this would be a good default or not.

[cppreference]: https://en.cppreference.com/w/c/types/limits
[godbolt-example]: https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:10,positionColumn:1,positionLineNumber:10,selectionStartColumn:1,selectionStartLineNumber:10,startColumn:1,startLineNumber:10),source:'%23include+%3Cstdint.h%3E%0A%0A%23if+SIZE_MAX%3D%3D0xffffffff%0A%23error+32bit+detected%0A%23elif+SIZE_MAX%3D%3D0xffffffffffffffff%0A%23error+64bit+detected%0A%23else%0A%23error+nothing+detected%0A%23endif%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang1701,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B11',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+17.0.1+(Editor+%231)',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+clang+17.0.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+17.0.1+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

zcbenz:
Thanks for the reference, according to MSDN this trick should work on msvc.

> SIZE_MAX | same as _UI64_MAX if _WIN64 is defined, or UINT_MAX

However for other platforms the general definition of `SIZE_WIDTH` is `bit width of object of size_t type`, the value of which completely depends on the implementation.

On the other hand `_WIN32` and `_WIN64` are well documented and there is no room for surprises.

zcbenz:
@aklomp Putting the quest for a generic macro define aside, can you merge this PR and do a patch release first?  Node.js updated base64 dep very recently and I would like to fix the broken build soon.

aklomp:
Hi everyone, sorry for not weighing in on this issue, but unfortunately my availability is very limited right now. I hope to resolve this in one or two weeks at most.

zcbenz:
Ping @aklomp for a review, I'm good with https://github.com/aklomp/base64/pull/125 too since it fixes my problem anyway.

aklomp:
@zcbenz Thanks for your patience. I merged #125 instead of this PR because that fix looks a bit more portable by not doing platform detection, and you mentioned that it also solves this issue.

Let me know if you still need a point release. I can do one, but I'd prefer to fix some other issues first (mainly #127).

zcbenz:
Thanks for looking in this, I'll just wait for a new release so I can update the dep in Node.js.

PR #125, opened 2023-11-19
fix: Detect platform word size based on `SIZE_MAX`

BurningEnlightenment:
The native word size on a platform matches the size of `size_t` with a few rare exceptions like the x32 ABI on linux. Therefore we can derive a general cross platform value for `BASE64_WORDSIZE` based on `SIZE_MAX` which the C99 standard guarantees to be available and well defined.

This fixes the misdection of x86/arm32 win32 as a 64bit platform, i.e. it is an alternative to #123 
@mayeut I think this should also cover the `__WORDSIZE` and `__SIZE_WIDTH__` cases. Can you verify this? If that indeed is the case I would like remove those branches, too.

See also https://en.cppreference.com/w/c/types/limits

### DIFF
--- a/lib/env.h
+++ b/lib/env.h
@@ -1,6 +1,8 @@
 #ifndef BASE64_ENV_H
 #define BASE64_ENV_H
 
+#include <stdint.h>
+
 // This header file contains macro definitions that describe certain aspects of
 // the compile-time environment. Compatibility and portability macros go here.
 
@@ -46,12 +48,14 @@
 #if defined (__x86_64__)
 // This also works for the x32 ABI, which has a 64-bit word size.
 #  define BASE64_WORDSIZE 64
-#elif defined (_INTEGRAL_MAX_BITS)
-#  define BASE64_WORDSIZE _INTEGRAL_MAX_BITS
 #elif defined (__WORDSIZE)
 #  define BASE64_WORDSIZE __WORDSIZE
 #elif defined (__SIZE_WIDTH__)
 #  define BASE64_WORDSIZE __SIZE_WIDTH__
+#elif SIZE_MAX == 0xffffffff
+#  define BASE64_WORDSIZE 32
+#elif SIZE_MAX == 0xffffffffffffffff
+#  define BASE64_WORDSIZE 64
 #else
 #  error BASE64_WORDSIZE_NOT_DEFINED
 #endif


lib/env.h
```diff
@@ -46,12 +48,14 @@
 #if defined (__x86_64__)
 // This also works for the x32 ABI, which has a 64-bit word size.
 #  define BASE64_WORDSIZE 64
-#elif defined (_INTEGRAL_MAX_BITS)
-#  define BASE64_WORDSIZE _INTEGRAL_MAX_BITS
 #elif defined (__WORDSIZE)
 #  define BASE64_WORDSIZE __WORDSIZE
 #elif defined (__SIZE_WIDTH__)
 #  define BASE64_WORDSIZE __SIZE_WIDTH__
+#elif SIZE_MAX == 0xffffffff
```
mayeut, line 55:
```suggestion
#elif SIZE_MAX == UINT32_MAX
```

lib/env.h
```diff
@@ -46,12 +48,14 @@
...
+#  define BASE64_WORDSIZE 32
+#elif SIZE_MAX == 0xffffffffffffffff
```
mayeut, line 57:
```suggestion
#elif SIZE_MAX == UINT64_MAX
```

mayeut:
@BurningEnlightenment, I did not write anything related to `__WORDSIZE`. For  `__SIZE_WIDTH__`, it was directly as bit-size for Alpine but indeed, using SIZE_MAX is equivalent.

BurningEnlightenment:
@mayeut sorry w.r.t. the `__WORDSIZE` misattribution--the git-blame was a bit misleading due to 955e9addde55d0fb4d1c0a56cb5a95c48c5bb16a. Apparently the `__WORDSIZE` usage goes all the way back to 41b1b7750d1a809b6ee9da0650622112f8fbd165. So I will go ahead and remove those branches.

aklomp:
Merged with the fixes proposed by @mayeut.

@BurningEnlightenment I slightly modified the commit title to add the issue number, and I modified the commit body to mark this issue and #123 as resolved. Hope you don't mind.

BurningEnlightenment:
@aklomp I additionally wanted to remove these four lines https://github.com/aklomp/base64/blob/88607feedaa57abd382d6ceb4b296629c6626f08/lib/env.h#L51-L54
should I open a new PR?

aklomp:
@BurningEnlightenment I think it's a bit too late now to force-push to the master branch, so yes, please open a new PR.

Issue #127, opened 2023-11-26
Build of 0.5.1 broken with MinGW


SpaceIm:
Build of 0.5.1 with MinGW11.0.0/GCC13 fails (it was working fine in 0.5.0):

```
[7/15] Building C object CMakeFiles/base64-bin.dir/bin/base64.c.obj
FAILED: CMakeFiles/base64-bin.dir/bin/base64.c.obj
D:\Programmes\mingw64\bin\gcc.exe  -IC:/Users/spaceim/.conan2/p/b/base6763a9b6f78112/b/src/include -m64 -O3 -DNDEBUG -MD -MT CMakeFiles/base64-bin.dir/bin/base64.c.obj -MF CMakeFiles\base64-bin.dir\bin\base64.c.obj.d -o CMakeFiles/base64-bin.dir/bin/base64.c.obj -c C:/Users/spaceim/.conan2/p/b/base6763a9b6f78112/b/src/bin/base64.c
C:/Users/spaceim/.conan2/p/b/base6763a9b6f78112/b/src/bin/base64.c:11:10: fatal error: sys/uio.h: No such file or directory
   11 | #include <sys/uio.h>
      |          ^~~~~~~~~~~
```

`sys/uio.h` is not provided by MinGW.

aklomp:
Indeed. The `base64` demo utility has been rewritten to use `writev(2)`,  which is defined in that header. Does MinGW supply that POSIX function somewhere else?

SpaceIm:
AFAIK, there is no `writev` defined in MinGW.

SpaceIm:
In conan recipe of base64, I've proposed to always disable build of this executable: https://github.com/conan-io/conan-center-index/pull/21386

sergeevabc:
Windows 7 x64 with SSE3 Core2Duo user here. I am glad to see that this issue has already been covered. Could someone be so kind to generate a dirty .exe and attach here as zip, while the team is looking for a general solution?

aklomp:
I pushed a branch called [issue127](https://github.com/aklomp/base64/tree/issue127) which emulates `writev(2)` on MinGW. I'd be happy to hear if someone can test it properly. I did test the emulation on Linux (by manually defining `EMULATE_WRITEV`) and it seems to work OK.

sergeevabc:
@aklomp, thank you for taking a moment to revisit this issue.
[...]

Issue #131, opened 2024-01-06
build failure using mingw32: `error: 'asm' operand has impossible constraints`


mayeut:
The assembly code fails to build on mingw32:
```
cmake -B out -Werror=dev -DBASE64_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release -DBASE64_WITH_AVX512=OFF
...
cmake --build out --config Release --verbose
Run Build Command(s): D:/a/_temp/msys64/mingw32/bin/ninja.exe -v
[1/18] D:\a\_temp\msys64\mingw32\bin\cc.exe -DBASE64_STATIC_DEFINE -ID:/a/base64/base64/include -ID:/a/base64/base64/out -O3 -DNDEBUG -std=c99 -Wall -Wextra -Wpedantic -Werror -mssse3 -MD -MT CMakeFiles/base64.dir/lib/arch/ssse3/codec.c.obj -MF CMakeFiles\base64.dir\lib\arch\ssse3\codec.c.obj.d -o CMakeFiles/base64.dir/lib/arch/ssse3/codec.c.obj -c D:/a/base64/base64/lib/arch/ssse3/codec.c
FAILED: CMakeFiles/base64.dir/lib/arch/ssse3/codec.c.obj 
D:\a\_temp\msys64\mingw32\bin\cc.exe -DBASE64_STATIC_DEFINE -ID:/a/base64/base64/include -ID:/a/base64/base64/out -O3 -DNDEBUG -std=c99 -Wall -Wextra -Wpedantic -Werror -mssse3 -MD -MT CMakeFiles/base64.dir/lib/arch/ssse3/codec.c.obj -MF CMakeFiles\base64.dir\lib\arch\ssse3\codec.c.obj.d -o CMakeFiles/base64.dir/lib/arch/ssse3/codec.c.obj -c D:/a/base64/base64/lib/arch/ssse3/codec.c
In file included from D:/a/base64/base64/lib/arch/ssse3/codec.c:29:
In function 'enc_loop_ssse3',
    inlined from 'base64_stream_encode_ssse3' at D:/a/base64/base64/lib/arch/ssse3/codec.c:42:2:
D:/a/base64/base64/lib/arch/ssse3/enc_loop_asm.c:160:9: error: 'asm' operand has impossible constraints
  160 |         __asm__ volatile (
      |         ^~~~~~~
[...]

Issue #103, opened 2022-08-24
Library naming


BurningEnlightenment:
During the review of the vcpkg integration of this library (microsoft/vcpkg#25091) it was noted that the name of this library is very generic and will likely clash with other base64 libraries. This is especially dangerous for names resolved by the linker like the so/archive name and function names. The public header name is also prone to a file name conflict.

Therefore it would be nice if either the library name was changed to something more unique like a proper name or the entities above were prefixed with something unique. Obviously this would be a breaking change and I would understand if this was rejected. OTOH it has been mentioned that 0.6.0 would be the target for breaking changes anyway.

aklomp:
This relates to #67. In that issue, I remarked that I consider it a reasonable thing for a `base64` library to claim the `base64` namespace. I am still partial to that opinion, although I also understand where you're coming from. Claiming a widely used prefix is not very social behavior when it comes to the kind of multi-tenant environment that vcpkg seems to be working in.

You're correct that I'm considering making some breaking changes for the next release. (Though if I do so, that release will probably be `1.0.0rc1` or `0.99.x` and not `0.6.0`, to indicate through the version number that there is an API discontinuity.) In that process, I would really like to overhaul pretty much everything in this library, and maybe renaming the library might be a nice crown on top of that work. No better break from the old than to start fresh under a new name.

The question remains, if we're going to rename it, what would be a good name? I'd like to include `base64` or `b64` somewhere in the name so that the library is easy to find.

aklomp:
I'm tending towards `akbase64` as a pretty short and reasonably memorable name.

BurningEnlightenment:
Sounds good to me 👍

kenhys:
Are there any progress?

In another point of cmake view,
Currently we need to use `find_package(base64)` in CMakeLists.txt but it occupies generic `base64` name.
That was claimed that it might be conflicted with other packages in debian.
so to avoid conflicts, added aklomp- prefix for it.
https://packages.debian.org/sid/amd64/libaklomp-base64-dev/filelist

It is just a workaround that distribution patches vendor specific one.
I hope that will be fixed in upstream too.

PR #132, opened 2024-01-06
CI: add tests for various MSYS2 environments

mayeut:
This PR adds CI tests for various MSYS2 environments (msys, mingw64, clang32 & clang64).

It's based on top of 17a9ab9b4ad1e6bf759282788745619a2abe5c0a and also adds a fix for mingw makefile builds.

Resolves https://github.com/aklomp/base64/issues/127.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -169,3 +169,46 @@ jobs:
             echo "::group::CTest"
             ctest --no-tests=error --test-dir out -VV --build-config Release
             echo "::endgroup::CTest"
+
+  msys2-cmake-test:
+    name: msys2-cmake-${{ matrix.msystem }}
+    needs: makefile-analysis
+    runs-on: windows-2022
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { msystem: msys, toolchain: "gcc" }
+          - { msystem: mingw32, env: mingw-w64-i686- }
+          - { msystem: mingw64, env: mingw-w64-x86_64- }
+          - { msystem: ucrt64, env: mingw-w64-ucrt-x86_64- }
+          - { msystem: clang32, env: mingw-w64-clang-i686- }
+          - { msystem: clang64, env: mingw-w64-clang-x86_64- }
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup MSYS2 ${{matrix.msystem}}
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{matrix.msystem}}
+          update: true
+          install: >-
+            make
+            ${{ matrix.env }}${{ matrix.toolchain || 'toolchain' }}
+            ${{ matrix.env }}cmake
+      - name: CMake Configure
+        shell: msys2 {0}
+        run: >
+          cmake
+          -B out
+          -Werror=dev
+          -DBASE64_BUILD_TESTS=ON
+          -DCMAKE_BUILD_TYPE=Release
+          -DBASE64_WITH_AVX512=OFF
+      - name: CMake Build
+        shell: msys2 {0}
+        run: cmake --build out --config Release --verbose
+      - name: CTest
+        shell: msys2 {0}
+        run: ctest --no-tests=error --test-dir out -VV --build-config Release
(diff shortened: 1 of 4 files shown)


aklomp:
Nice, thanks! I think there's a dependency on #131 though; it would be nice to fix that one first so that we don't need to comment out the 32-bit builds.

mayeut:
I could [check](https://github.com/mayeut/base64/actions/runs/7437256198/job/20234700625) that #125 fixes the 32-bit build.
It would require one more commit here to pass the makefile check as all symbols are prefixed by `_`. I'll add this commit.

mayeut:
@aklomp, if you're ok with the changes here, I think the next steps are either:
- merge this one, rebase #123 / #125 & remove the comment on 32-bit mingw
- merge #123 / #125, rebase this one &  remove the comment on 32-bit mingw

mayeut:
@aklomp, I rebased on master and removed the comment on ming 32-bit.

test/Makefile
```diff
@@ -15,7 +17,9 @@ endif
 
 test: clean test_base64 benchmark
 	./test_base64
+ifeq (, $(findstring mingw, $(TARGET)))  # no "/dev/urandom" on Windows
```
aklomp, line 20: There's a really old pull request, #6, which fixes this on Windows. Now that we have a provisional CI environment for Windows, I can try to revive that branch and see if I can get the benchmark to work.

Having a working Windows benchmark would also be great for checking in CI that certain changes did not introduce speed regressions on that platform.

Makefile
```diff
@@ -64,9 +66,16 @@ all: bin/base64 lib/libbase64.o
...
 
-lib/libbase64.o: $(OBJS)
-	$(LD) -r -o $@ $^
-	$(OBJCOPY) --keep-global-symbols=lib/exports.txt $@
+lib/exports.build.txt: lib/exports.txt
+ifeq (i686-w64-mingw32, $(TARGET))
+	sed -e 's/^/_/' $< > $@
+else
+	cp -f $< $@
+endif
```
aklomp, line 74: Honestly I'm not a big fan of this compatibility block, though I understand why it's needed. In the longer term I think we should try to get rid of the exports file entirely and rely on compiler/linker features to mark functions as hidden or visible.

BurningEnlightenment: You mean like annotating these symbols as `BASE64_SYMBOL_PRIVATE`?

.github/workflows/test.yml
```diff
@@ -212,3 +212,34 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { msystem: msys, toolchain: "gcc" }
+          # - { msystem: mingw32, env: mingw-w64-i686- }  disabled, c.f. https://github.com/aklomp/base64/issues/131
+          - { msystem: mingw64, env: mingw-w64-x86_64- }
+          - { msystem: ucrt64, env: mingw-w64-ucrt-x86_64- }
+          # - { msystem: clang32, env: mingw-w64-clang-i686- }  disabled, lld does not support the "-r" option
+          # - { msystem: clang64, env: mingw-w64-clang-x86_64- }  disabled, lld does not support the "-r" option
```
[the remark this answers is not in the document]
mayeut: No. The feature is supported for ELF & Mach-O targets but not for Windows.

Makefile
mayeut: I don't like this either but if we're to keep the current semantic, it's needed.
It's not the same as `BASE64_SYMBOL_PRIVATE`:
`BASE64_SYMBOL_PRIVATE` only acts on the visibility attribute of `GLOBAL` symbols e.g. even if `base64_stream_encode_plain` is marked as hidden it's still linkable (if building a static library or bare object).
`--keep-global-symbols` acts on `LOCAL`/`GLOBAL` attribute. It effectively marks symbols that are not in the list as `LOCAL` & those are not linkable anymore (as if we only had 1 translation unit - simulated by `ld -r` - and the function was declared `static`).

I wouldn't mind if the makefile produced the same things as the CMake build does for a static library (just something like `ar -o libbase64.a *.o`, getting rid of the `ld -r` and `objcopy`).

Another option is not to support makefile builds for Windows at all.

aklomp: My idea for the future roadmap is basically to engineer the library such that all functions start with `akbase64_` (which is probably what I will rename this library to, see #103), *and* that all functions are declared `static`, except for those that are part of the public API. Those two measures should remove all potential naming conflicts in practice.

Until then, I suggest we keep going as usual, so using this Makefile hack/complication out of necessity.

PR #133, opened 2024-01-08
refactor: Remove detection logic subsumed by `SIZE_MAX` checks

BurningEnlightenment:
@aklomp as per https://github.com/aklomp/base64/pull/125#issuecomment-1881039242


aklomp:
Mind if I reword the commit so it starts with the issue slug and contains the magic phrase `Resolves #133.` at the bottom?

BurningEnlightenment:
I'm fine with that :)

PR #143, opened 2024-06-12
fix: Correct OS detection for AVX-512

### LINKED ISSUE #113: bug: codec_choose_x86 does not check for OS AVX512 support
The check for OS support has not been updated & still only checks for AVX support.
AVX512 added masks, 512-bit register size & 16 new registers. There are 3 corresponding bits that need to be checked in [XCR0](https://en.wikipedia.org/wiki/Control_register#XCR0_and_XSS).

mgaudet:
This fixes crashes for me locally 

Fixes #113

### DIFF
--- a/lib/codec_choose.c
+++ b/lib/codec_choose.c
@@ -74,7 +74,16 @@
 #define _XCR_XFEATURE_ENABLED_MASK 0
 #endif
 
-#define _XCR_XMM_AND_YMM_STATE_ENABLED_BY_OS 0x6
+#define bit_XMM      (1 << 1)
+#define bit_YMM      (1 << 2)
+#define bit_OPMASK   (1 << 5)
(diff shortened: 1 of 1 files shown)


lib/codec_choose.c
```diff
@@ -74,7 +74,16 @@
 #define _XCR_XFEATURE_ENABLED_MASK 0
 #endif
 
-#define _XCR_XMM_AND_YMM_STATE_ENABLED_BY_OS 0x6
+#define bit_XMM      (1 << 1)
+#define bit_YMM      (1 << 2)
+#define bit_OPMASK   (1 << 5)
+#define bit_ZMM      (1 << 6)
+#define bit_HIHG_ZMM (1 << 7)
```
mgaudet, line 81: whoops tyop

PR #150, opened 2025-04-23
feat: add NEON support on Windows ARM64

mayeut:
add NEON support on Windows ARM64

NEON is mandatory on aarch64/arm64, checking for HAVE_NEON64 is enough, no need for other preprocessing macros.


aklomp:
Thanks for the PR and sorry for the delay in handling it. I'll try to get to this soon. Looks good to me at first glance.

PR #157, opened 2026-07-12
fix(ci): update actions & runners

mayeut:
Update all actions used in the test workflow.
ubuntu-22.04, macos-13, macos-14 and windows-2019 are deprecated or already removed => update to ubuntu-24.04, macos-15 & windows-2022.
With the update to node 24 to run GitHub Actions, Alpine 3.12 can't be run as a container => update to Alpine 3.21 which is the oldest not EOL alpine version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,13 @@
 name: Test
 
-on: [push, pull_request]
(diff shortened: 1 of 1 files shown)


aklomp:
Merged, thanks!

